PR #1, opened 2020-11-16
adding semantic-release action

dlockhart:
Adds a shared action for releasing our components (and other projects) using the `semantic-release` project.

### DIFF
--- a/semantic-release/README.md
+++ b/semantic-release/README.md
@@ -4,7 +4,7 @@ This GitHub action uses the [semantic-release](https://semantic-release.gitbook.
 * Release notes will be generated
 * The version in `package.json` will be incremented
 * A GitHub Release will be created
-* The package will be published to NPM
+* Optionally, the package will be published to NPM
 
 ## Using the Action
 
@@ -34,13 +34,25 @@ jobs:
         uses: BrightspaceUI/actions/semantic-release@master
         with:
           GITHUB_TOKEN: ${{ secrets.SPECIAL_GITHUB_TOKEN }}
+          NPM: true
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
 Notes:
 * If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Semantic Release" step.
 * In the checkout step, you must set the `persist-credentials` option to `false`. This opts out of the default `GITHUB_TOKEN` which is not an admin and cannot bypass branch protection rules.
-* `NPM_TOKEN` is available automatically as a shared organization secret in the `BrightspaceUI`, `BrightspaceUILabs` and `BrightspaceHypermediaComponents` organizations.
+
+## NPM Package Deployment
+
+If you'd like the action to deploy your package to NPM, set the `NPM` option to `true` and pass through the `NPM_TOKEN` secret. `NPM_TOKEN` is available automatically as a shared organization secret in the `BrightspaceUI`, `BrightspaceUILabs` and `BrightspaceHypermediaComponents` organizations.
+
+If your package is being published under the `@brightspace-ui` or `@brightspace-ui-labs` NPM organizations, ensure that it has the proper configuration in its `package.json`:
+
+```json
+"publishConfig": {
+  "access": "public"
+}
+```
 
 ## Triggering a Release
 
--- a/semantic-release/README.md
+++ b/semantic-release/README.md
@@ -40,6 +40,12 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 ```
 
+Options:
+* `DRY_RUN` (default: `false`): Runs semantic-release with the `--dry-run` flag to simulate a release but not actually do one
+* `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create GitHub release
+* `NPM` (default: `false`): Whether or not to release as an NPM package
+* `NPM_TOKEN`: Token to publish to NPM
+
 Notes:
 * If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Semantic Release" step.
 * In the checkout step, you must set the `persist-credentials` option to `false`. This opts out of the default `GITHUB_TOKEN` which is not an admin and cannot bypass branch protection rules.
@@ -56,6 +62,8 @@ If your package is being published under the `@brightspace-ui` or `@brightspace-
   "access": "public"
 }
 ```
+
+Also ensure that `"private": true` is not present.
 
 ## Triggering a Release
 
--- a/semantic-release/README.md
+++ b/semantic-release/README.md
@@ -43,8 +43,8 @@ jobs:
 Options:
 * `DRY_RUN` (default: `false`): Runs semantic-release with the `--dry-run` flag to simulate a release but not actually do one
 * `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create GitHub release
-* `NPM` (default: `false`): Whether or not to release as an NPM package
-* `NPM_TOKEN`: Token to publish to NPM
+* `NPM` (default: `false`): Whether or not to release as an NPM package (see "NPM Package Deployment" below for more info)
+* `NPM_TOKEN` (optional if `NPM` is `false`): Token to publish to NPM (see "NPM Package Deployment" below for more info)
 
 Notes:
 * If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Semantic Release" step.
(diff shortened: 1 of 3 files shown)


semantic-release/README.md
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+        with:
+          persist-credentials: false
+      - name: Setup Node
+        uses: Brightspace/third-party-actions@actions/setup-node
+        # additional validation steps can be run here
+      - name: Semantic Release
+        uses: BrightspaceUI/actions/semantic-release@master
```
dlockhart, line 34: `@master` is what we want right?

semantic-release/README.md
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          GITHUB_TOKEN: ${{ secrets.SPECIAL_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+```
+
+Notes:
+* If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Semantic Release" step.
+* In the checkout step, you must set the `persist-credentials` option to `false`. This opts out of the default `GITHUB_TOKEN` which is not an admin and cannot bypass branch protection rules.
```
dlockhart, line 42: I was planning on adding more documentation here and fixing the example once we actually know what the final approach will be.

semantic-release/README.md
```diff
@@ -0,0 +1,60 @@
+# Semantic Release Action
+
+This GitHub action uses the [semantic-release](https://semantic-release.gitbook.io/) project to automatically analyze commits since the previous release and increment the package version according to [semantic versioning](https://semver.org/) rules. If it's determined that a release should be created:
+* Release notes will be generated
+* The version in `package.json` will be incremented
+* A GitHub Release will be created
+* The package will be published to NPM
```
dlockhart, line 7: I'm wondering if deployment to NPM should be optional...

semantic-release/README.md
svanherk: Yup that's correct

semantic-release/README.md
svanherk: I think that would help teams move over easier.  I found the config setup needed to look like this if you want to turn off npm publish (but still commit the version bump to the `package.json` file):
```
"release": {
    "plugins": [
      "@semantic-release/commit-analyzer",
      "@semantic-release/github",
      [
        "@semantic-release/npm",
        {
          "npmPublish": false
        }
      ],
      "@semantic-release/release-notes-generator",
      [
        "@semantic-release/git",
        {
          "assets": [
            "package.json"
          ],
          "message": "chore(release): ${nextRelease.version} [skip ci]"
        }
      ]
    ]
  }
```

svanherk: We could make the NPM_TOKEN optional and if they pass it in create one config, and if not, the other.

semantic-release/README.md
```diff
@@ -0,0 +1,60 @@
...
+        uses: BrightspaceUI/actions/semantic-release@master
+        with:
+          GITHUB_TOKEN: ${{ secrets.SPECIAL_GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+```
+
+Notes:
+* If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Semantic Release" step.
+* In the checkout step, you must set the `persist-credentials` option to `false`. This opts out of the default `GITHUB_TOKEN` which is not an admin and cannot bypass branch protection rules.
+* `NPM_TOKEN` is available automatically as a shared organization secret in the `BrightspaceUI`, `BrightspaceUILabs` and `BrightspaceHypermediaComponents` organizations.
```
svanherk, line 43: Interesting... for some reason I thought this had to be a repo-specific token?

semantic-release/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    required: true
+  NPM_TOKEN:
+    description: Token to publish to NPM
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Installing semantic-release
+      run: |
+        if [ ${{ !contains(github.event.head_commit.message, 'skip ci') }} == true ]; then
```
svanherk, line 15: Would this work here?
```
if [ ${{ !contains(github.event.head_commit.message, 'skip ci') }} == true ]; then
          echo "Installing semantic-release..."
          npm install semantic-release@17 @semantic-release/git@9
else
          echo "[skip ci] detected so skipping release"
          exit 0;
fi
```
And then we don't need this check in the steps below?  Or do they continue even after exiting?

semantic-release/README.md
dlockhart: Nope, we can now have one token that the whole org (or all orgs) share.

semantic-release/action.yml
dlockhart: I tried that earlier -- it does not. It exits _that step_ but continues with the rest.

semantic-release/README.md
dlockhart: Cool -- I think I'll just have an `NPM` true/false option that toggles between the two config files. Will update this PR.

semantic-release/action.yml
svanherk: Right, because we don't fail the step.  Based on what I'm seeing I think it would work if we did `exit 1`, but I don't think we want to _fail_ skipped builds... hmm.

svanherk: Apparently there used to be a neutral exit code that _may_ have done what we wanted, but it's not supported anymore: https://github.community/t/github-actions-neutral-exit-code-is-incorrectly-interpreted-as-failure/16088/7

dlockhart: Yeah exactly. And it still kind-of sucks because in the Actions UI it shows up as "passed" whereas if it used an `if` in the workflow it has a greyed out icon indicating that it was skipped. But hopefully they add `if` support to composite actions soon and we can fix this.

svanherk: Yeah, it's probably still best practice for us to add it to the `if` of the deploying workflows as well, to cut off all the steps before it.  But then if someone forgets, this catches us

semantic-release/README.md
dlockhart: Ok added!

semantic-release/action.yml
```diff
@@ -0,0 +1,45 @@
...
+inputs:
+  GITHUB_TOKEN:
+    description: Token to use to update version in 'package.json' and create GitHub release
+    required: true
+  NPM:
+    description: Whether or not to release as an NPM package
+    default: false
+  NPM_TOKEN:
+    description: Token to publish to NPM
+    required: true
```
svanherk, line 12: I don't think we want this required anymore

semantic-release/create-config.sh
```diff
@@ -0,0 +1,25 @@
+NPM_STEP=""
+if [ "$NPM" = "true" ]
+then
+  NPM_STEP=$'\n    "@semantic-release/npm",'
```
svanherk, line 4: Does excluding this completely still up the version in `package.json`?  I couldn't get that to work before when I tried it

dlockhart: Hmm you're right, it doesn't appear to be doing that anymore. `@semantic-release/npm` has an option to do the publish or not, so I'll just toggle that on or off. Good catch!

semantic-release/README.md
```diff
@@ -0,0 +1,93 @@
...
+          GITHUB_TOKEN: ${{ secrets.SPECIAL_GITHUB_TOKEN }}
+          NPM: true
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+```
+
+Options:
+* `DRY_RUN` (default: `false`): Runs semantic-release with the `--dry-run` flag to simulate a release but not actually do one
+* `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create GitHub release
+* `NPM` (default: `false`): Whether or not to release as an NPM package
+* `NPM_TOKEN`: Token to publish to NPM
```
svanherk, line 47: Nice, I like the dry run addition!  Do we want to call out `NPM_TOKEN` as optional?  It's maybe implied by the existence of  `NPM`

dlockhart: Cool yeah agreed -- added some more info to the option docs.

svanherk (approved):
This looks good to me!  Once the token is available, we can set that up and then mention here how to get `SECRET_GITHB_TOKEN` (or whatever we call it) to be automatically available for their repo

PR #2, opened 2020-11-17
incremental release action

dlockhart:
This action recreates the release portion of [frau-ci](https://github.com/Brightspace/frau-ci) with a GitHub action.

It can be used in places where using semantic-release may be overkill or not appropriate, such as in non-libraries which just want to release every time something merges.

### DIFF
--- a/incremental-release/README.md
+++ b/incremental-release/README.md
@@ -16,6 +16,7 @@ on:
       - master
 jobs:
   release:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
     timeout-minutes: 2
     runs-on: ubuntu-latest
     steps:
@@ -33,7 +34,7 @@ jobs:
 ```
 
 Options:
-* `DEFAULT_INCREMENT` (default: `skip`): if no release keyword is found in the latest commit message, this value will be used to trigger a release
+* `DEFAULT_INCREMENT` (default: `skip`): If no release keyword is found in the latest commit message, this value will be used to trigger a release. Can be one of: `skip`, `patch`, `minor`, `major`.
 * `DRY_RUN` (default: `false`): Simulates a release but does not actually do one
 * `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create the tag
 
@@ -47,6 +48,8 @@ Releases occur based on the most recent commit message:
 * Commits which contain `[increment patch]` will trigger a `patch` release. Example: `validate input before using [increment patch]`
 * Commits which contain `[increment minor]` will trigger a `minor` release. Example: `add toggle() method [increment minor]`
 * Commits which contain `[increment major]` will trigger a `major` release. Example: `breaking all the things [increment major]`
+
+**Note:** When merging a pull request, this will be the merge commit message.
 
 ### Default Increment
 
--- a/incremental-release/README.md
+++ b/incremental-release/README.md
@@ -1,0 +1,65 @@
+# Incremental Release Action
+
+This GitHub action looks for keywords in the latest commit to automatically increment the package version and create a tag.
+
+## Using the Action
+
+Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge.
+
+Here's a sample release workflow:
+
+```yml
+name: Release
+on:
+  push:
+    branches:
+      - master
+jobs:
+  release:
+    timeout-minutes: 2
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+        with:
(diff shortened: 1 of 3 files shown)


incremental-release/README.md
```diff
@@ -0,0 +1,65 @@
...
+Here's a sample release workflow:
+
+```yml
+name: Release
+on:
+  push:
+    branches:
+      - master
+jobs:
+  release:
```
svanherk, line 18: Should we put the skip ci stuff in the readme (and in the semantic-release one as well)?  In case people copy/paste?

incremental-release/README.md
```diff
@@ -0,0 +1,65 @@
...
+        uses: Brightspace/third-party-actions@actions/setup-node
+        # additional validation steps can be run here
+      - name: Incremental Release
+        uses: BrightspaceUI/actions/incremental-release@master
+        with:
+          GITHUB_TOKEN: ${{ secrets.D2L_GITHUB_TOKEN }}
+```
+
+Options:
+* `DEFAULT_INCREMENT` (default: `skip`): if no release keyword is found in the latest commit message, this value will be used to trigger a release
```
svanherk, line 36: Should we list all possible values here?  Not too hard to find in the code but means they wouldn't need to

svanherk: I guess it's _kinda_ in the `Triggering a Release` or `Default Increment` section

incremental-release/README.md
```diff
@@ -0,0 +1,65 @@
...
+* `DRY_RUN` (default: `false`): Simulates a release but does not actually do one
+* `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create the tag
+
+Notes:
+* If you have additional release validation steps (e.g. build step, validation tests), run them after the "Setup Node" step and before the "Incremental Release" step.
+* In the checkout step, you must set the `persist-credentials` option to `false`. This opts out of the default `GITHUB_TOKEN` which is not an admin and cannot bypass branch protection rules.
+
+## Triggering a Release
+
+Releases occur based on the most recent commit message:
```
svanherk, line 46: Should we call out here that this is most likely to be the merge commit?  An image of where to update it probably overkill, but I have personally sent one to three different people to help explain why things weren't working 😛

incremental-release/README.md
svanherk: Hmm, maybe this interferes with `[skip release]`.  Although technically, both should skip it and did in Travis

incremental-release/action.yml
```diff
@@ -0,0 +1,63 @@
...
+        else
+          echo "Did not find commit message flag, using DEFAULT_INCREMENT ($DEFAULT_INCREMENT)"
+        fi
+
+        if [ $TYPE == "skip" ]; then
+          echo "Skipping release."
+          exit 0;
+        fi
+
+        git config user.name github-actions
```
svanherk, line 45: This works even when we're not using `GITHUB_TOKEN` right?  This doesn't need to match whoever the access token is for?  I don't think I've ever verified that...

incremental-release/action.yml
```diff
@@ -0,0 +1,63 @@
...
+        if [ ${{ inputs.DRY_RUN }} == true ]; then
+          echo "DRY_RUN option specified, aborting release creation."
+          exit 0;
+        fi
+        
+        echo "Executing: \"git push --follow-tags\""
+        git push --follow-tags
+
+      env:
+        DEFAULT_INCREMENT: ${{ inputs.DEFAULT_INCREMENT }}
```
svanherk, line 61: Interested in why this needed to be an environment variable instead of being accessed like `DRY_RUN` is

incremental-release/README.md
dlockhart: I think it's worth adding -- added.

incremental-release/README.md
dlockhart: Added.

incremental-release/README.md
dlockhart: Haha yeah it's confusing -- added.

incremental-release/action.yml
dlockhart: I don't believe so, no. Because in my testing, the token is for the brightspace-bot, which is _not_ the github-actions user.

incremental-release/action.yml
dlockhart: Good catch. So it was more about how I was using them differently in the bash script. `$DEFAULT_INCREMENT` would have needed to be an environment variable, but `${{ inputs.DEFAULT_INCREMENT }}` uses the input. I've adjusted it to just use the input for both.

svanherk (approved):
Looks good to me! 🎉

PR #3, opened 2020-11-18
Add prepare-translations-for-npm action to core actions

grant-cleary:
This is my go at converting [this CLI script](https://github.com/BrightspaceUI/core/blob/master/cli/verify-translations.js) into a reusable action. The major difference here is that instead of blowing up if we don't find one of the expected language files, we instead just ignore it and move on.

I'm not super convinced I'm implementing this as well as it could be implemented, so I'm 100% open to suggestions!


prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,43 @@
+# Prepare Translations for NPM Action
```
grant-cleary, line 1: I don't know that I'm very good at writing `README`s and I have no idea if this is too much information or not enough. This started based on a copy/paste of the semantic-release action's README to try to align sections as much as possible.

prepare-translations-for-npm/prepare-translations.js
```diff
@@ -0,0 +1,57 @@
...
+			Object.keys(translations).sort().forEach((key) => {
+				ordered[key] = translations[key];
+			});
+			_writeChanges(langTermsPath, lang, ordered);
+		}
+	});
+}
+
+try {
+	prepare(process.env['LANG_PATH']);
```
grant-cleary, line 51: This is effectively the same as the existing CLI script, except we take the lang path in as a parameter and ignore any missing translation files. I think if we want to throw some kind of warning for missing translation files, it's maybe more appropriate for that to happen in some CI-based verify step (as discussed in Slack), but I'm open to opinions otherwise.

prepare-translations-for-npm/action.yml
```diff
@@ -0,0 +1,18 @@
...
+    description: Path to your repo's translation files
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Installing chalk
+      run: npm install --prefix ${{ github.action_path }} chalk@4
+      shell: bash
+    - name: Preparing translations
+      run: node ${{ github.action_path }}/prepare-translations.js
```
grant-cleary, line 14: This is a bit gross, but from what I can tell, it's not really possible to invoke JS directly in composite actions, so if we want to install chalk, we have to do it this way.

prepare-translations-for-npm/action.yml
```diff
@@ -0,0 +1,18 @@
...
+      env:
+        LANG_PATH: ${{ github.workspace }}/${{ inputs.LANG_PATH }}
+        FORCE_COLOR: 3
```
grant-cleary, line 17: Turns out chalk needs this set for the console to allow colours.

prepare-translations-for-npm/prepare-translations.js
```diff
@@ -0,0 +1,57 @@
+const chalk = require('chalk'),
+	fs = require('fs');
+
+const langs = ['ar', 'cy', 'da', 'de', 'es', 'es-es', 'fr', 'fr-fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh', 'zebras'];
```
grant-cleary, line 4: Oh dear. This was in there for testing...

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,43 @@
+# Prepare Translations for NPM Action
+
+This GitHub action prepares translation files in a project for NPM publishing by creating temporary `en` translations for each language term and each language
+in your project that doesn't yet have a proper translation.
+
+## Using the Action
+
+Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge.
```
dbatiste, line 8: Is that true?  Wouldn't we typically run it only when we are about to release?

prepare-translations-for-npm/prepare-translations.js
dbatiste: That's fair - if we did want to try and handle the missing language case in a more automatic way, I think it would need to happen in a different job anyway, because it would need to commit a couple changes (the new file, the serge config file, and we'd need to share the localize mixin somehow). An action that verifies could be the beginning of that.

dbatiste (approved):
Seems ok to me, but I'd wait for Dave or Stacey to also give a review since I'm not really that familiar with the GHA specifics.

prepare-translations-for-npm/README.md
grant-cleary: I believe so. If this isn't followed by a publish to NPM, it's sort of useless. It makes changes to the translation files in the GHA workspace, but doesn't touch anything in the repo proper.

svanherk: Maybe for clarity we say:
"Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge that you are planning to publish to npm (for example, before running the `semantic-release` action <link to https://github.com/BrightspaceUI/actions/tree/master/semantic-release>)."

grant-cleary: This seems good to me.

margaree (approved):

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          LANG_PATH: './lang'
+        # NPM publishing step happens here
+```
+
+Options:
+* `LANG_PATH`: The relative path to your language files from the root of your repository
+
+Notes:
+* If you have additional release validation steps (e.g. build step, validation tests), run them before the "Prepare Translations for NPM" step.
```
dlockhart, line 42: I think this bullet could be excluded, since it's really more for the release actions.

prepare-translations-for-npm/action.yml
```diff
@@ -0,0 +1,18 @@
+name: Prepare Translations For NPM
+description: Prepares translations for NPM publishing, ensuring that new lang terms have temporary English translations
+inputs:
+  LANG_PATH:
+    description: Path to your repo's translation files
+    required: true
```
dlockhart, line 6: Wondering if there's a common directory that most repos use that we could default to? e.g. `lang`

prepare-translations-for-npm/action.yml
```diff
@@ -0,0 +1,18 @@
...
+description: Prepares translations for NPM publishing, ensuring that new lang terms have temporary English translations
+inputs:
+  LANG_PATH:
+    description: Path to your repo's translation files
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Installing chalk
+      run: npm install --prefix ${{ github.action_path }} chalk@4
```
dlockhart, line 11: You'll want the `--no-save` flag, otherwise it'll add it to `package.json` as a dependency and the semantic-release action will check that into source control.

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,43 @@
+# Prepare Translations for NPM Action
+
+This GitHub action prepares translation files in a project for NPM publishing by creating temporary `en` translations for each language term and each language
+in your project that doesn't yet have a proper translation.
+
+## Using the Action
+
+Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge that you are planning to publish to npm (for example, before running the [`semantic-release` action](https://github.com/BrightspaceUI/actions/tree/master/semantic-release).
```
svanherk, line 8:
```suggestion
Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge that you are planning to publish to npm (for example, before running the [`semantic-release` action](https://github.com/BrightspaceUI/actions/tree/master/semantic-release)).

```

grant-cleary: Ah, thanks for the catch!

prepare-translations-for-npm/prepare-translations.js
```diff
@@ -0,0 +1,57 @@
+const chalk = require('chalk'),
+	fs = require('fs');
+
+const langs = ['ar', 'cy', 'da', 'de', 'es', 'es-es', 'fr', 'fr-fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
+
+function _parseFile(filePath) {
```
dlockhart, line 6: Wondering if we could experiment here with importing the file as a JavaScript module?

```javascript
const lang = await import(filePath);
return lang;
```

Not sure if dynamic imports are actually supported from Node.js though.

prepare-translations-for-npm/README.md
grant-cleary: I mostly left it in because the "Prepare Translations" step does modify the working directory, which means tests technically won't be getting a "clean" checkout to work on. I don't know whether that matters in practice or whether it's common to write tests around translations. I'll happily yield to your experience here.

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,44 @@
+# Prepare Translations for NPM Action
+
+This GitHub action prepares translation files in a project for NPM publishing by creating temporary `en` translations for each language term and each language
+in your project that doesn't yet have a proper translation.
+
+## Using the Action
+
+Typically this action is triggered from a workflow that runs on your `master` branch after each commit or pull request merge that you are planning to publish to npm (for example, before running the [`semantic-release` action](https://github.com/BrightspaceUI/actions/tree/master/semantic-release)).
```
grant-cleary, line 8:
```suggestion
Typically this action is called from a release workflow before a step which publishes to NPM, such as the [semantic-release action](https://github.com/BrightspaceUI/actions/tree/master/semantic-release).
```

grant-cleary: Apparently I can't commit suggestions based on "outdated" diffs, so this is just your suggestion re-applied, @dlockhart.

prepare-translations-for-npm/action.yml
grant-cleary: Oh really? Yikes, will fix. 😬

grant-cleary, line 11:
```suggestion
      run: npm install --no-save --prefix ${{ github.action_path }} chalk@4
```

svanherk: Were you seeing this in your actions output?  I think adding this should also fix that (it did for me):
```
npm WARN saveError ENOENT: no such file or directory, open '/home/runner/work/_actions/svanherk/test-action/main/visual-diff/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/home/runner/work/_actions/svanherk/test-action/main/visual-diff/package.json'
```

prepare-translations-for-npm/prepare-translations.js
```diff
@@ -0,0 +1,57 @@
...
+	});
+}
+
+try {
+	prepare(process.env['LANG_PATH']);
+	process.exit(0);
+} catch (err) {
+	console.error(chalk.red(err));
+	console.groupEnd();
+	process.exit(1);
```
dlockhart, line 56: Were you able to confirm that exiting the JS script in this way actually fails the action and stops a deployment (subsequent steps from running)?

prepare-translations-for-npm/README.md
dlockhart: That's fair -- I'm good either way, more info is usually better!

prepare-translations-for-npm/prepare-translations.js
grant-cleary: I did confirm that this fails the action appropriately. I didn't explicitly try with something occurring later, though I based this on the GHA documentation implying that anything != 0 is a failure state. I hope it wouldn't continue to run later steps, but I can verify that.

prepare-translations-for-npm/action.yml
grant-cleary: I'm less familiar with this, but it feels like defaulting to `lang` would be okay. Worst case scenario, the action fails.

prepare-translations-for-npm/prepare-translations.js
```diff
@@ -0,0 +1,57 @@
+const chalk = require('chalk'),
+	fs = require('fs');
+
+const langs = ['ar', 'cy', 'da', 'de', 'es', 'es-es', 'fr', 'fr-fr', 'ja', 'ko', 'nl', 'pt', 'sv', 'tr', 'zh-tw', 'zh'];
+
+function _parseFile(filePath) {
+	const file = fs.readFileSync(filePath).toString();
+	const firstBit = file.split('default ')[1];
```
svanherk, line 8: I'm wondering if we should call out the format of the languages files in a separate section in the README.  This is expecting our new JS format of
```

/* eslint quotes: 0 */

export default {
	"components.alert.close": "Close Alert",
};
```
and is not going to work for anyone using the old JSON files setup.

prepare-translations-for-npm/prepare-translations.js
dlockhart: Cool seems safe to assume that. We can always test it out once this gets integrated into core or the HTML editor workflows.

prepare-translations-for-npm/prepare-translations.js
dlockhart: Agreed yeah, this only works with the JavaScript module format.

prepare-translations-for-npm/prepare-translations.js
grant-cleary: I planned to integrate this into the HTML editor immediately where the cost of a potential failure is low, so I'll definitely want to re-test it once that's done. :)

prepare-translations-for-npm/prepare-translations.js
grant-cleary: Ah yeah, good point. Do we want this to work with the JSON format, or do we want to just call out that this only works with JS modules?

prepare-translations-for-npm/prepare-translations.js
grant-cleary: Good question. I'm trying to find out which Node version dynamic imports were introduced in, but it might be better/easier just to add it, test it, and find out.

prepare-translations-for-npm/action.yml
grant-cleary: Ah yeah, I did see that. I assumed this was just something NPM had to do on a temporary basis so I didn't think much of it.

prepare-translations-for-npm/prepare-translations.js
svanherk: The JS module way is _much_ cleaner and doesn't require a build step like the JSON method did.  The migration from JSON to JS is also not too bad, so I say skip it and only support JS

grant-cleary: Maybe a `Limitations` section outlining that this can't be used with the old JSON format? Or does `Format` with a less negative type of message make more sense?

dlockhart: Agreed that we don't need to support the JSON format. It also can't be unbundled, so it's a performance problem. We want people moving off of it.

prepare-translations-for-npm/prepare-translations.js
grant-cleary: Okay, so I forked and did some testing and it appears to work, modulo it can't actually `await` because the action calls this script synchronously. But yeah, dynamic imports appear to work just fine here, at least with Node v12.19.0 (what I used in my test). Is there some minimum version of Node we should be supporting? I can redo my testing with that number.

prepare-translations-for-npm/prepare-translations.js
grant-cleary: I also confirmed that this failure does block further job steps.

prepare-translations-for-npm/prepare-translations.js
svanherk: I think something like `Lang File Format`

prepare-translations-for-npm/prepare-translations.js
dlockhart: Looks like it was added in version 12.

grant-cleary: Makes sense. Is demanding Node 12 reasonable? I think 12 is the current LTS according to [their release page](https://nodejs.org/en/about/releases/), but if we have reasons we don't want to push/enforce that, I can leave it as is.

svanherk: Node 12 seems reasonable, I think we use that in most of our Travis jobs.  But we'd have to rely on the consumer installing that, which is a bit odd but not crazy.  I think Dave L and I saw in some of his runs that is we didn't specify, it defaulted to 10

grant-cleary: Yeah, it's a bit of a weird workflow. I imagine I could add a step to the composite action to check for the correct Node version as well as put a note in the README. The latter seems sensible regardless, though I'm not sure if the former is overkill.

dlockhart: README should suffice I think. The [docs for the setup-node action](https://github.com/actions/setup-node) indicate that specifying a version is pretty easy. No idea how the default is resolved though.

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,55 @@
...
+* When setting up NodeJS, the version specified must be `>= 12`. This action uses features introduced in v12.
+* If you have additional release validation steps (e.g. build step, validation tests), run them before the "Prepare Translations for NPM" step.
+* You must run the checkout step before running the "Prepare Translations for NPM" step, otherwise the action will not be able to find your lang path.
+* Don't run this action if you're not publishing to NPM. It doesn't commit to your repo and isn't relevant outside of publishing to NPM.
+
+## Lang File Format
+
+This action expects the following format for lang files:
+* All lang files should be contained within the top level of the `LANG_PATH` directory. This action does not support language files within nested directories (e.g. `{LANG_PATH}/es/es.js`, `{LANG_PATH}/en/en.js` would not be supported).
+* Lang files must be JavaScript modules, formatted as outlined in the [localize mixin documentation](https://github.com/BrightspaceUI/core/blob/master/mixins/localize-mixin.md).
```
grant-cleary, line 53: I'm unsure if we have better documentation on this process, but if we do, I would be glad to link it here (or I can remove this reference if we think it's implied).

grant-cleary: I also added some extra info here that I thought was relevant (that empty default exports need to be present for supported langs, that lang files should go into the main list, etc.). I can remove or change them if there are objections.

svanherk: So if we add a new language to the list, and they don't have a js file for it, does it fail the build with a helpful message?  Or we pass and warn?  Or we pass?

grant-cleary: We pass. Because this workflow is meant to come on release (post-build), it's challenging to get a message to the consumer. We could put it in the build log, but if the build passes, I'm not sure people will read it. I'm definitely open to having this work a different way, but I figured we would want to leave build failures out of this action in favour of (at some point) introducing a CI-level action that would actually fail the build if lang files were missing.

svanherk: Yeah I think that's fine, just wasn't totally apparent from the wording and wanted to add info about failing if we did so.

grant-cleary: Ah, if it's not apparent from the wording, then I should make it so. I'll add some clarification there.

svanherk: Awesome, that feels a lot clearer to me now

dlockhart (approved):

prepare-translations-for-npm/README.md
```diff
@@ -0,0 +1,53 @@
...
+    branches:
+      - master
+jobs:
+  release:
+    timeout-minutes: 2
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+        with:
```
svanherk, line 26:
```suggestion
```

grant-cleary: Is omitting `with:` valid when inputs are provided? I guess there's no explicit need to specify `persist-credentials` in this example workflow.

svanherk: Oh maybe I did the wrong `with`

grant-cleary: Oh, did I leave it in for the Node piece? Whoops.

svanherk: Yeah sorry, meant to put it on that one 🤦 

I don't mind having the `persist-credentials` here even though they don't need it for this, because they probably _would_ need it for the npm publish one (if they use our action)

grant-cleary: Yeah, I added it back and removed the extraneous `with`. I reconsidered and we may as well encourage use of our release actions.

grant-cleary:
I'm away for the rest of this week, so if someone wants to make use of this before next Monday, please feel free to merge it on my behalf. Otherwise, I'm happy to resolve any merge conflicts, review comments, etc. and merge it when I'm back next week. 😄

svanherk (approved):
Looks great to me, thanks @grant-cleary!

grant-cleary:
Haha well that also works! I haven't left yet, so I'll go ahead and merge. Thanks all for being patient with me as I stumbled through this!

PR #4, opened 2020-11-26
using a script to create a release

dlockhart:
This addresses a couple issues:

First, the actual `push` was failing to use the elevated GitHub token. Switching it to include the token in the repo URL seems to have solved things.

Second, `frau-ci` created GitHub releases and this wasn't doing that, so I've added a JavaScript helper that uses the Octokit API to do this.

### DIFF
--- a/incremental-release/action.yml
+++ b/incremental-release/action.yml
@@ -73,10 +73,6 @@ runs:
           echo "DRY_RUN option specified, skipping release creation."
           exit 0;
         fi
-        if [ ${{ steps.increment-version.outputs.version }} == "" ]; then
-          echo "No new version, skipping release creation."
-          exit 0;
-        fi
         node ${{ github.action_path }}/create-release.js ${{ steps.increment-version.outputs.version }}
       env:
         GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
--- a/incremental-release/action.yml
+++ b/incremental-release/action.yml
@@ -13,15 +13,21 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Incremental Release
+    - name: Install Dependencies
       run: |
+        if [ ${{ !contains(github.event.head_commit.message, 'skip ci') }} == true ]; then
+          echo "Installing dependencies..."
+          npm install @octokit/rest@18 --prefix ${{ github.action_path }} --no-save --loglevel error
+        fi
+      shell: bash
+    - name: Determine Increment
+      id: determine-increment
+      run: |
+        TYPE=${{ inputs.DEFAULT_INCREMENT }}
         if [ ${{ contains(github.event.head_commit.message, 'skip ci') }} == true ]; then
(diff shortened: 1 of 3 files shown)


incremental-release/action.yml
```diff
@@ -36,27 +42,42 @@ runs:
...
-          echo "DRY_RUN option specified, aborting release creation."
+          echo "DRY_RUN option specified, skipping release creation."
           exit 0;
         fi
-        
-        echo "Executing: \"git push --follow-tags\""
-        git push --follow-tags
-
+        if [ ${{ steps.increment-version.outputs.version }} == "" ]; then
+          echo "No new version, skipping release creation."
```
svanherk, line 77: Were you able to test out this case?  What would cause the `npm` version step to fail?

incremental-release/create-release.js
```diff
@@ -0,0 +1,33 @@
+const { Octokit } = require('@octokit/rest');
+
+const release = process.argv[2];
+const [owner, repo] = process.env['GITHUB_REPOSITORY'].split('/');
+
+const octokit = new Octokit({
+	auth: process.env.GITHUB_TOKEN
```
svanherk, line 7: I wonder if we actually need `persist-credentials: false` for this action... since you use the input directly in the git push call, and specifically set the auth here.  I don't think it'll use the config at all, which was the problem before... 🤔 
If so, we could remove that from the README

incremental-release/create-release.js
```diff
@@ -0,0 +1,33 @@
...
+	console.log(`Creating release "${release}"...`)
+
+	const response = await octokit.repos.createRelease({
+		owner: owner,
+		repo: repo,
+		tag_name: release,
+		name: release
+	});
+	if (response.status !== 201) {
+		console.error(response);
```
svanherk, line 21: If this happens, do we want to fail the build?

incremental-release/action.yml
dlockhart: Yeah I can't get it to be hit... we don't call `npm version` if it's not going to bump. I'll remove this check.

incremental-release/create-release.js
dlockhart: I'll give it a try!

incremental-release/create-release.js
dlockhart: I'll try to trigger this again... it may actually be throwing in which case it's not needed.

incremental-release/action.yml
dlockhart: Removed.

incremental-release/create-release.js
dlockhart: Refactored this as the call to `createRelease` throws, so it's not needed.

incremental-release/create-release.js
dlockhart: You were correct, removed.

svanherk (approved):

PR #7, opened 2020-12-10
Visual-diff action enhancements

svanherk:
After trying to use it in a few places, the following enhancements were discovered:
- Don't install `visual-diff` and `puppeteer` if we're in the `visual-diff` repo (necessary, not really an enhancement lol)
- Include untracked files when staging (ran into this issue locally when I was creating only new files, no updates)
- Check if the goldens branch exists before trying to delete it, to avoid unnecessary error messages when the build is green
- Adding more details to the goldens PR description to make it clearer what the PR is for
- Adding a comment back on the original PR pointing to the goldens PR

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -25,7 +25,10 @@ runs:
     - name: Install Dependencies
       run: |
         echo -e "\e[34mInstalling Dependencies"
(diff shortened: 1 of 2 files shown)


dlockhart (approved):
Looks great!

PR #8, opened 2020-12-10
Adding RELEASE and VERSION outputs

dlockhart:
Was about to quickly convert `ifrau` and realized that it needs to know whether a release occurred or not and if it did what the new version is. This is needed by `frau-publisher` for CDN publishing, so most FRAs will need the same.

This outputs those two values.

### DIFF
--- a/semantic-release/action.yml
+++ b/semantic-release/action.yml
@@ -16,6 +16,9 @@ outputs:
   RELEASE:
     description: Whether or not a release occurred
     value: ${{ steps.semantic-release.outputs.release }}
+  VERSION:
+    description: Version of the new release
+    value: ${{ steps.semantic-release.outputs.version }}
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


semantic-release/action.yml
```diff
@@ -45,7 +55,15 @@ runs:
...
+          OLD_VERSION=$(node -p -e "require('./package.json').version")
           echo "Running semantic-release..."
           npx semantic-release
+          NEW_VERSION=$(node -p -e "require('./package.json').version")
+          if [ "$OLD_VERSION" == "$NEW_VERSION" ]; then
+            echo "::set-output name=release::false"
+          else 
+            echo "::set-output name=release::true"
+            echo "::set-output name=version::$(echo $NEW_VERSION)"
+          fi
```
svanherk, line 67: Could this just be:
```
          NEW_VERSION=$(node -p -e "require('./package.json').version")
          if [ "$OLD_VERSION" != "$NEW_VERSION" ]; then 
            echo "::set-output name=release::true"
            echo "::set-output name=version::$(echo $NEW_VERSION)"
          fi
```
Since you set it to `false` already above?

svanherk (approved):

semantic-release/action.yml
dlockhart: Indeed -- thx!

PR #10, opened 2020-12-15
excluding node_modules from default mocha test path

dlockhart:
Repos like `consistent-evaluation` that include other D2L projects with visual-diff tests using a GitHub reference run into a scenario where the default test path (`./**/*.visual-diff.js`) finds the visual diff tests in those dependency repos.

So we either change the default to just look in `./test/**/*.visual-diff.js` where we'd need to fix a few repos like `core`, or we apply this fix which will _exclude_ `node_modules`.


dbatiste (approved):

dlockhart:
I'll merge this for now. @svanherk if you want to take the opposite approach when you get back, it should be an easy change.

svanherk:
I'm good with this change!  Thanks!

PR #13, opened 2021-01-07
Add warning when the stash apply fails because of merge conflicts from outdated goldens

svanherk:
I tried numerous ways to avoid this scenario - by applying the goldens using stashes, by resetting, by trying to fix the merge conflicts automatically, etc.  In the end, GitHub and git really don't handle binary file conflicts well and would still complain about a conflict that needs to be resolved manually.

This tries to warn of the conflict early - if your branch is missing changes to golden files that you will be changing, the visual-diff tests fail and a comment is made on your PR telling you to rebase or merge in master (or whatever branch is the PR base).  However, if someone merges their golden changes _after_ you've already updated the goldens in your branch, you'll get the conflict.  This would have happened regardless.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -82,7 +82,11 @@ runs:
         git fetch origin +refs/heads/$SOURCE_BRANCH:refs/heads/$SOURCE_BRANCH -q || true
         git checkout $SOURCE_BRANCH
         git checkout -b $VISUAL_DIFF_BRANCH
-        git stash apply
+        if ! git stash apply; then
+          echo -e "\e[31mCould not apply stash - merge conflicts with ${{ github.event.pull_request.base.ref }}."
+          echo "::set-output name=goldens-conflict::$(echo true)"
+          exit 0;
+        fi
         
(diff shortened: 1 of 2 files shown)


visual-diff/action.yml
```diff
@@ -98,17 +102,23 @@ runs:
       shell: bash
     - name: Handling the Goldens PR (if necessary)
       run: |   
-        if [ ${{ steps.visual-diff-tests.outputs.tests-passed }} == true ]; then
+        if [ ${{ steps.visual-diff-tests.outputs.tests-passed }} == true ] || [ ${{ steps.committing-goldens.outputs.goldens-conflict }} == true ]; then
```
svanherk, line 105: If a goldens PR is already up and CI reruns and finds the conflict, I figured we might as well close the old PR since merging it would cause problems.  We could also leave it open and it will update after the rebase, if we want to avoid a new PR being created (more emails and slack notifications), but I thought it'd be better to close the invalid PR asap.

dlockhart (approved):
Looks good!

PR #14, opened 2021-01-11
Quit Out of Runs on Stale Commits

svanherk:
Margaree ran into this issue when merging a visual-diff PR while the tests were still running on another commit in the PR.  This could happen a lot in `core` since the test runs take a while, and I couldn't find a setting to make GitHub Actions just cancel themselves.  So now, after running the visual-diff tests, we check if there are any new commits that have been made to our branch in the meantime before trying to put up or modify the goldens PR.  If there have been, we just stop the workflow run - we don't want to be changing anything based on old data, and that new commit would've triggered another visual-diff workflow to start.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -64,9 +64,11 @@ runs:
         if [ ${{ github.event.number }} ]; then
           SOURCE_BRANCH=${{ github.event.pull_request.head.ref }}
           VISUAL_DIFF_BRANCH=ghworkflow/visual-diff-pr-${{ github.event.number }}
+          ORIGINAL_SHA=${{ github.event.pull_request.head.sha }}
         else
           SOURCE_BRANCH=${GITHUB_REF#refs/heads/}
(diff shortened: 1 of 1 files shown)


visual-diff/action.yml
```diff
@@ -79,9 +81,15 @@ runs:
...
-        git fetch origin +refs/heads/$SOURCE_BRANCH:refs/heads/$SOURCE_BRANCH -q || true
+        git fetch origin +refs/heads/$SOURCE_BRANCH:refs/heads/$SOURCE_BRANCH -q -u || true
         git checkout $SOURCE_BRANCH
         git checkout -b $VISUAL_DIFF_BRANCH
+        if [ $(git rev-parse HEAD) != $ORIGINAL_SHA ]; then
+          echo -e "\e[31mBranch out of date - more commits have been added to the '$SOURCE_BRANCH' branch since this action started running.  Stopping this test run."
+          exit 1;
+        fi
+        
+        echo "::set-output name=goldens-conflict::$(echo false)"
```
svanherk, line 92: This is somewhat unrelated - fixing a bash error when this wasn't set.

dlockhart (approved):

PR #15, opened 2021-01-13
Add visual-diff branch cleanup

svanherk:
This was a feature we talked about that was a nice-to-have, so was deferred until after we'd gotten Travis replaced everywhere.

At the beginning of the action, we pull all the visual diff PR branches and make sure the PR they correspond to is still open.  If not, the branch is deleted.  This helps cleanup branches if a PR was closed but the corresponding visual-diff one was not - if the PR was later reopened, visual-diff would have rerun anyways.

It would make more sense to run this on every PR close but that would require it's own workflow, or on a schedule but that feels like overkill.  Running it every time we run visual-diff tests is a compromise.

### DIFF
--- a/visual-diff/cleanup-branches.js
+++ b/visual-diff/cleanup-branches.js
@@ -3,17 +3,60 @@
 const chalk = require('chalk');
 const { Octokit } = require('@octokit/rest');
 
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Nice!

PR #20, opened 2021-02-17
docs: Add info about assuming hub roles

svanherk:
Adding info about assuming the role if using a hub role.  I wasn't sure if I should add it to the main example, or do what I did below.  I'm also going to put up a PR to update https://github.com/Brightspace/iam-build-tokens/blob/master/docs/howto-githubactions.md#using-the-tokens, but given the GitHub Actions doc is separated from the Hub Role doc, I think it could easily be missed.


dlockhart (approved):
👍

PR #23, opened 2021-03-25
Fix config

AntonBazhal:


semantic-release/create-config.sh
```diff
@@ -6,16 +6,14 @@ else
...
-      {name: "alpha", prerelease: true}
-    ]
-  },
+  "branches": [
+    "+([0-9])?(.{+([0-9]),x}).x",
+    "$DEFAULT_BRANCH",
+    "next",
+    "next-major",
+    {"name": "beta", "prerelease": true},
+    {"name": "alpha", "prerelease": true}
```
AntonBazhal, line 15: this part was not a proper JSON (no "")

semantic-release/create-config.sh
```diff
@@ -6,16 +6,14 @@ else
...
-    "branches": [
-      "+([0-9])?(.{+([0-9]),x}).x",
-      "$DEFAULT_BRANCH",
-      "next",
-      "next-major",
-      {name: "beta", prerelease: true},
-      {name: "alpha", prerelease: true}
-    ]
-  },
+  "branches": [
```
AntonBazhal, line 9: `branches` does not need to be under `release` if it's not a `package.json`, so this config was not taking effect at all

dlockhart (approved):
I have no idea how this was working before... maybe it was inferring that it was JSON despite the `yml` extension. Anyway, thanks!

PR #24, opened 2021-04-15
fix(semantic-release): resolve variable value.

neverendingqs:
I noticed a `/home/runner/work/_actions/BrightspaceUI/actions/master/semantic-release/create-config.sh: line 2: [: !=: unary operator expected
` error on https://github.com/Brightspace/frau-locale-provider/runs/2353432455#step:4:19. I think this resolves it.

(Sorry not sure how to test this change).

### DIFF
--- a/semantic-release/create-config.sh
+++ b/semantic-release/create-config.sh
@@ -1,5 +1,5 @@
 IS_TRACKED=$(git ls-files package-lock.json || true)
-if [ $IS_TRACKED != "" ]; then
+if [ "$IS_TRACKED" != "" ]; then
   ASSETS="[\"package.json\", \"package-lock.json\"]"
 else
   ASSETS="[\"package.json\"]"


semantic-release/create-config.sh
```diff
@@ -1,5 +1,5 @@
 IS_TRACKED=$(git ls-files package-lock.json || true)
-if [ $IS_TRACKED != "" ]; then
+if [ "$IS_TRACKED" != "" ]; then
```
neverendingqs, line 2: If I'm reading https://stackoverflow.com/questions/13617843/unary-operator-expected-error-in-bash-if-condition correctly, this should resolve this error:

```
/home/runner/work/_actions/BrightspaceUI/actions/master/semantic-release/create-config.sh: line 2: [: !=: unary operator expected
```

semantic-release/create-config.sh
```diff
@@ -1,5 +1,5 @@
 IS_TRACKED=$(git ls-files package-lock.json || true)
```
neverendingqs, line 1: Out-of-scope, but valuable for catching errors:

```suggestion
set -eu
IS_TRACKED=$(git ls-files package-lock.json || true)
```
`-e` ~= throw on error and `-u` ~= error on unset variable (https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html).

dlockhart: Weirdly doing this in my test action didn't cause anything to fail:
https://github.com/dlockhart/semantic-release-action-test/runs/2353894542?check_suite_focus=true

semantic-release/create-config.sh
dlockhart: This seems to solve the error!

semantic-release/create-config.sh
neverendingqs: I'm kinda really confused now. I ran this locally using Git Bash without issues (note the lack of `"`s around `${IS_TRACKED}`):

`set-e-iff.sh`: 
```bash
#!/bin/bash

set -eu

IS_TRACKED=a
if [ $IS_TRACKED != "" ]; then
  ASSETS="[\"package.json\", \"package-lock.json\"]"
else
  ASSETS="[\"package.json\"]"
fi

echo ${ASSETS}
```

```bash
$ ./set-e-iff.sh
["package.json", "package-lock.json"]
```

The action says to use bash... https://github.com/BrightspaceUI/actions/blob/abfe9e5cb06ff48b74cd1bd6490a22b4f7630839/semantic-release/action.yml#L32-L42

Maybe it's related to bash version or OS?

```bash
$ bash --version
GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

neverendingqs: This will fail, so I'm guessing errors inside `if`s don't count for `set -e`?

```bash
#!/bin/bash

set -eu
grep a b
```

```bash
$ ./set-e-iff.sh
grep: b: No such file or directory
```

neverendingqs: I'm comfortable merging as-is or with `set -eu` added. I think there's value in adding the `"`s.

dlockhart: Yeah I definitely feel like this was working at one point in the not-so-distant past. Will merge as-is.

dlockhart (approved):

PR #26, opened 2021-04-19
Visual Diff Action Enhancements

svanherk:
This PR:
1. Bumps the version of `puppeteer` and `visual-diff`
2. Installs and uses `esm` in the mocha call, to enable teams to write their visual-diff tests as an ES module and to enable us to eventually switch `visual-diff` to an ES module.
3. Fixes up an error message that was showing up as a red herring a lot.  We now differentiate between a general error and a stash conflict error, and comment on the PR accordingly.  This is related to https://github.com/BrightspaceUI/core/pull/1183#issuecomment-790748400.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -27,7 +27,9 @@ runs:
         echo -e "\e[34mInstalling Dependencies"
         NPM_PACKAGE=$(cut -d "=" -f 2 <<< $(npm run env | grep "npm_package_name"))
         if [ $NPM_PACKAGE != '@brightspace-ui/visual-diff' ]; then
-          npm install mocha@8 puppeteer@7 @brightspace-ui/visual-diff@3 --no-save
+          npm install esm@3 mocha@8 puppeteer@8 @brightspace-ui/visual-diff@4 --no-save
+        else
+          npm install esm@3 --no-save
         fi
         npm install chalk@4 @octokit/rest@18 --prefix ${{ github.action_path }} --no-save --loglevel error
       env:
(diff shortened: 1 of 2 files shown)


visual-diff/action.yml
```diff
@@ -27,7 +27,9 @@ runs:
         echo -e "\e[34mInstalling Dependencies"
         NPM_PACKAGE=$(cut -d "=" -f 2 <<< $(npm run env | grep "npm_package_name"))
         if [ $NPM_PACKAGE != '@brightspace-ui/visual-diff' ]; then
-          npm install mocha@8 puppeteer@7 @brightspace-ui/visual-diff@3 --no-save
+          npm install esm@3 mocha@8 puppeteer@8 @brightspace-ui/visual-diff@4 --no-save
+        else
+          npm install esm@3 --no-save
```
svanherk, line 32: If I just did this line by itself after the `if` statement, it didn't work (puppeteer is somehow uninstalled - it all needs to be installed together)

visual-diff/action.yml
```diff
@@ -89,7 +91,13 @@ runs:
           exit 1;
         fi
         
+        echo "::set-output name=changes-empty::$(echo false)"
         echo "::set-output name=goldens-conflict::$(echo false)"
+        if [ $(git stash list | wc -l) == 0 ]; then
```
svanherk, line 96: We check here if no stash was created - that means we had some general issue (can't find any test files, test import doesn't exist, etc).  If there is a stash, but we can't apply, _then_ we should complain about a version conflict with `master`.

dbatiste (approved):

GeorgePlukov (approved):

krshnpatel (approved):
This is great! Thanks 😃

PR #27, opened 2021-04-22
Switch the goldens PR to open as a draft PR

svanherk:
Opening the goldens PR as a draft PR means teams won't get notified about it in slack (if drafts are ignored) or as codeowners until it's actually ready to be merged, meaning less noise about flakes and bugs found and fixed.  It will mean that you'll now need to press "Ready for review" before merging the PR.

This will only open the PR as a draft.  If you change it to a "Ready to review" PR and then more golden changes are added, it won't turn it back into a draft PR.


dlockhart (approved):
Amazing -- thanks!

PR #31, opened 2021-07-07
Update frau publisher example

mostlyfabulous:
As per the args syntax:
```
frau-publisher --moduletype|-m app 
               --targetdir|-t 'cdn directory' 
               --key|-k yourkey 
               --secret|-s yoursecret 
               --devtag yourtag 
               --version|-v yourversion ex. 0.0.1 
               --files|-f './dist/**'
```

I also missed that `--devtag` must be passed: https://github.com/Brightspace/frau-publisher/blob/master/src/optionsValidator.js#L41


dlockhart (approved):
Cool, thanks! I think in the places I had used this it was only for production publishing not dev so the devtag hadn't been required.

PR #32, opened 2021-08-04
Add update-package-lock action

svanherk:
We'd talked about making an action out of the BSI update workflow (https://github.com/Brightspace/brightspace-integration/blob/master/.github/workflows/bsi-update.yml) for a while, so I figured I'd tackle it while diving into the graphql stuff for automerging anyways.  I'll be putting up a BSI PR shortly to switch the workflow over to use this action.

### DIFF
--- a/update-package-lock/README.md
+++ b/update-package-lock/README.md
@@ -1,0 +1,56 @@
+# Update `package-lock.json` Action
+
+This GitHub action will determine if any updates can be made to your repo's `package-lock.json`. If so, a PR will be opened with the updates against the repo's default release branch. You can also configure this action to auto-approve the PR for you, and set it to auto-merge after tests pass.
+
+## Using the Action
+
+Typically this action would be scheduled to run on whatever cadence works well for your repo. Our [brightspace-integration repo](https://github.com/Brightspace/brightspace-integration) runs it hourly to catch new changes across all our web components, whereas a regular project repo may only want to run it once a day or once a week.
+
+This action provides similar functionality to what dependabot gives you, but will combine all changes in one PR and can be set to automerge if all tests pass. It also will not change the `package.json` file, only handling updates that fall within the range you've set there. This means the action can provide a nice compliment to dependabot - it can handle all the "easy" updates for you automatically, so your time is mostly spent on reviewing/debugging major updates and breaking changes.
+
+Here's a sample workflow:
+
+```yml
+name: Update package-lock.json
+on:
+  schedule:
+    - cron: "0 15 * * 1-5" # Mon-Fri 11:00AM EDT. 10:00AM EST.
+jobs:
+  test:
+    timeout-minutes: 10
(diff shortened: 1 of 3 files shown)


update-package-lock/README.md
```diff
@@ -0,0 +1,56 @@
...
+  test:
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
+      - name: Update package-lock.json
+        uses: BrightspaceUI/actions/update-package-lock@master
+        with:
+          DEFAULT_BRANCH: master
```
svanherk, line 28: Figured I'd make the default `main` since no one is using this action yet, but add this in the example here while most people are still on `master`

update-package-lock/action.yml
```diff
@@ -0,0 +1,88 @@
+name: Update package-lock.json
+description: Opens a pull request with an updated package-lock.json
+inputs:
+  APPROVAL_TOKEN:
+    description: Token for automatically approving the PR (will skip if not provided)
+  AUTO_MERGE_TOKEN:
+    description: Token for enabling auto-merge for the PR (will skip if not provided)
```
svanherk, line 7: I figured I'd just use the existence of the token to determine whether to run these steps, rather than having a boolean and token for each

update-package-lock/action.yml
```diff
@@ -0,0 +1,88 @@
...
+    default: main
+  GITHUB_TOKEN:
+    description: Token for opening the pull request
+    required: true
+  PR_TITLE:
+    description: Title for the opened pull request
+    default: 'Updating package-lock.json'
+runs:
+  using: composite
+  steps:
```
svanherk, line 25: Most of this is taken from https://github.com/Brightspace/brightspace-integration/blob/master/.github/workflows/bsi-update.yml, and then I had to rewrite the `github-script` stuff since we can't use actions inside actions

update-package-lock/handle-pr.js
```diff
@@ -0,0 +1,134 @@
...
+			headers: {
+				authorization: `token ${autoMergeToken}`,
+			}
+		});
+
+		try {
+			await graphqlForAutoMerge(
+				`mutation enableAutoMerge($pullRequestId: String!) {
+					enablePullRequestAutoMerge(input: { pullRequestId: $pullRequestId }) {
+						clientMutationId
```
svanherk, line 88: I don't actually need the `clientMutationId` for anything, but I couldn't find a way to just ask for nothing

dlockhart (approved):
Looks awesome!

update-package-lock/README.md
dlockhart: Yep good call -- I want to eventually switch the default of the other actions around too but that's more involved.

PR #35, opened 2021-10-04
Switch to actual user name and email of github actions bot

devpow112:
You can find this from the users github api
https://api.github.com/users/github-actions%5Bbot%5D/events/public


dlockhart (approved):
Cool -- thanks!

PR #40, opened 2021-12-01
updating docs to refer to actions from "main" branch

dlockhart:
I've created a `main` branch off `master` and this PR just updates the documentation to refer to actions from main instead of master.

After this I'll switch the default branch to be `main`, which will leave `master` as a placeholder until we make some breaking changes and update everyone to point to main.

### DIFF
--- a/prepare-translations-for-npm/README.md
+++ b/prepare-translations-for-npm/README.md
@@ -5,7 +5,7 @@ in your project that doesn't yet have a proper translation.
 
 ## Using the Action
 
-Typically this action is called from a release workflow before a step which publishes to NPM, such as the [semantic-release action](https://github.com/BrightspaceUI/actions/tree/master/semantic-release).
+Typically this action is called from a release workflow before a step which publishes to NPM, such as the [semantic-release action](https://github.com/BrightspaceUI/actions/tree/main/semantic-release).
 
 
 Here's a sample workflow:
@@ -15,7 +15,7 @@ name: Release
 on:
   push:
     branches:
-      - master
+      - main
 jobs:
   release:
     timeout-minutes: 2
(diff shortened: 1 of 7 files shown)


prepare-translations-for-npm/README.md
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: Brightspace/third-party-actions@actions/setup-node
         # additional validation steps can be run here
       - name: Prepare Translations for NPM
-        uses: BrightspaceUI/actions/prepare-translations-for-npm@master
+        uses: BrightspaceUI/actions/prepare-translations-for-npm@main
         with:
           LANG_PATH: lang
         # NPM publishing step happens here
```
svanherk, line 35: Could update the link to the localize mixin below to use `main` while we're here

semantic-release/README.md
```diff
@@ -33,8 +33,9 @@ jobs:
       - name: Setup Node
         uses: Brightspace/third-party-actions@actions/setup-node
       - name: Semantic Release
-        uses: BrightspaceUI/actions/semantic-release@master
+        uses: BrightspaceUI/actions/semantic-release@main
         with:
+          DEFAULT_BRANCH: main
```
svanherk, line 38: We should update `semantic-release`'s default to be `main` for this version, and then when we move everyone over we can switch them

dlockhart: That's a great idea. So would we rename their branches for them, or just add in `DEFAULT_BRANCH: master`?

svanherk: I was thinking just add `DEFAULT_BRANCH: master` until they have time to get to it, to be less disruptive

dlockhart: Done.

svanherk (approved):

PR #41, opened 2021-12-01
ESM

dlockhart:
This makes the actions repo ESM-by-default and switches any JS files to use ESM import syntax. Upgrading to `chalk` version 5 which is ESM-only.


svanherk (approved):
Chalk 5 also requires node 12 - wonder if we should call that out on the main readme somewhere

PR #42, opened 2021-12-01
upgrading mocha, puppeteer and visual-diff

dlockhart:
We're now ready to use the latest Mocha@9, Puppeteer@12 and visual-diff@7.


svanherk (approved):
Woohoo!  Surprised I didn't have more instructions on the `esm` stuff but 🤷

PR #44, opened 2021-12-02
listing out versions action is using

dlockhart:
Something like this?


svanherk (approved):
Yup!  Wasn't sure if we actually wanted to say the `npm` line in case they don't know the `@` syntax, but I'm good either way

svanherk (approved):

PR #47, opened 2021-12-10
feat: allow .mjs extension

dlockhart:
There are certain repos that are just going to be too difficult to convert to be ESM-by-default. As an alternative, we can make the `*.visual-diff.js` files use the `.mjs` extension. That way Mocha will treat them as modules.

This change allows for that without teams needing to override the `TEST_PATH` themselves [like I did in the navigation repo(https://github.com/BrightspaceUI/navigation/blob/main/.github/workflows/visual-diff.yml#L22).


svanherk (approved):
Cool!

PR #48, opened 2021-12-15
removing from master branch

dlockhart:
BSI was the last place still pointing at the `master` branch for the `update-package-lock` action. Now that it's switched over, we can delete this action from the branch.

This is going to be our strategy for starving out the remaining actions -- once all the repos are switched over we'll remove it. Then once all the actions are gone we can delete the `master` branch.


annabelsegalld2l (commented):
Sorry, a bit confused - if we want to delete the action from the master branch, why is the PR pointing to the main branch?

dlockhart:
Lol you are justly confused! That was a mistake. Closing.

PR #51, opened 2021-12-20
Update Puppeteer in Visual Diff

svanherk:
So not sure if the action runs are still using 12, or if it was listening to the peer dependency and using 13... 😕


dlockhart (approved):
Right, forgot to bump this! Shouldn't have any visual impact.

PR #52, opened 2021-12-23
Remove prepare-translations-for-npm from master

svanherk:
Looks like everyone using this action is pointing to `main` now!


dlockhart (approved):
Awesome!

PR #57, opened 2022-02-16
Include CodeArtifact input to allow the action to read from CodeArtifact

ianmclean2011:


ianmclean2011:
Not sure about a default value. I assume the default is nothing, so not passing in a value should be fine (if it doesn't try to get something from CodeArtifact, it shouldn't need a token).

dlockhart (approved):
This LGTM -- we can test it quickly afterwards to confirm!

dlockhart (commented):
Oh there's also the README that lists out the options.

PR #61, opened 2022-03-02
Escaped the branch name

MitchelPaulin:
If a branch name contains a special character like a bracket it can cause the vdiff action to fail, see here https://github.com/Brightspace/d2l-rubric/pull/1691. Escaping the branch name should fix this problem. 

Note: This is a low risk change but I am not sure how I can test it locally, if anyone has suggestions I can try testing it before we merge this.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -66,7 +66,7 @@ runs:
       id: committing-goldens
       run: |
         if [ ${{ github.event.number }} ]; then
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Yeah it's pretty hard to test without forks and your own copies of the actions. I'm OK with merging this and then resolving any issues that might come up.

devpow112 (approved):

PR #63, opened 2022-03-09
Add `NODE_AUTH_TOKEN`

devpow112:
Will fallback to `CODEARTIFACT_AUTH_TOKEN` if not available. Working towards getting BSI to work with the standard private registry setup via the `setup-node` action (https://github.com/Brightspace/brightspace-integration/pull/5942).

### DIFF
--- a/update-package-lock/README.md
+++ b/update-package-lock/README.md
@@ -36,7 +36,8 @@ Options:
 * `APPROVAL_TOKEN`: Token to auto-approve the PR after opening. If no token is passed, the auto-approval will be skipped. See [setup details](#setting-up-auto-approval) below.
 * `AUTO_MERGE_TOKEN`: Token to enable auto-merge on the PR. If no token is passed, the auto-merge will be skipped. See [setup details](#setting-up-auto-merge) below.
 * `BRANCH_NAME` (default: `ghworkflow/package_lock_auto_update`): Name of the branch to add the changes to and open the pull request from.
-* `CODEARTIFACT_AUTH_TOKEN`: Token for reading @d2l packages from CodeArtifact. If no token is passed, any attempts at installing private packages from CodeArtifact will fail with a 401 error.
+* `NODE_AUTH_TOKEN`: Token for reading packages from already setup private registry
+* `CODEARTIFACT_AUTH_TOKEN`: Token for reading @d2l packages from CodeArtifact. If no token is passed, any attempts at installing private packages from CodeArtifact will fail with a 401 error (deprecated, use `NODE_AUTH_TOKEN` instead).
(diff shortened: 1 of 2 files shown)


update-package-lock/action.yml
```diff
@@ -25,6 +27,15 @@ inputs:
 runs:
   using: composite
   steps:
+      - name: Determine Token
```
devpow112, line 30: Can be removed once everyone is switched off using the `CODEARTIFACT_AUTH_TOKEN`

devpow112:
Is there anyway to test this other then push and hope for the best?

update-package-lock/README.md
```diff
@@ -36,7 +36,8 @@ Options:
 * `APPROVAL_TOKEN`: Token to auto-approve the PR after opening. If no token is passed, the auto-approval will be skipped. See [setup details](#setting-up-auto-approval) below.
 * `AUTO_MERGE_TOKEN`: Token to enable auto-merge on the PR. If no token is passed, the auto-merge will be skipped. See [setup details](#setting-up-auto-merge) below.
 * `BRANCH_NAME` (default: `ghworkflow/package_lock_auto_update`): Name of the branch to add the changes to and open the pull request from.
-* `CODEARTIFACT_AUTH_TOKEN`: Token for reading @d2l packages from CodeArtifact. If no token is passed, any attempts at installing private packages from CodeArtifact will fail with a 401 error.
+* `NODE_AUTH_TOKEN`: Token for reading packages from already setup private registry
+* `CODEARTIFACT_AUTH_TOKEN`: Token for reading @d2l packages from CodeArtifact. If no token is passed, any attempts at installing private packages from CodeArtifact will fail with a 401 error (deprecated, use `NODE_AUTH_TOKEN` instead).
```
dlockhart, line 40: Looks like this is [only used in 3 places](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=CODEARTIFACT_AUTH_TOKEN+%22actions%2Fupdate-package-lock%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5&si=full&si=full), so we could probably clean those up quickly after this merges and then remove it entirely.

dlockhart (approved):

update-package-lock/README.md
svanherk: And it looks like all those places are setting it as an environment variable, but not actually passing it into the action.  So I think this can be cleaned up now?

devpow112: I'll audit them today and get them fixed up along with this 👍. Was gonna do that Friday but ended up OOO unexpectedly

PR #64, opened 2022-03-14
Add ability to set merge method

devpow112:
Sometimes I don't want merge commits 😄.


svanherk:
Yeah I don't think I realized you could pick one of these when setting auto-merge, otherwise I definitely would've chosen `squash` 😝

PR #66, opened 2022-03-14
Add DEFAULT_BRANCH_PRERELEASE

AntonBazhal:
**The problem**
I'm trying to integrate `semantic-release` into the CI pipeline for our services. Currently, every merge to `master` gets deployed to the `dev` stage and we create a release when we want to deploy to `prod`. Unfortunately, we will end up with `prod` deployments on every merge by switching to `semantic-release`.

**The solution**
I'm totally open to other suggestions but my change allows us to create pre-releases on merge to `master` (with only `dev` deployment) and just marking them as "ready for release" when we want a `prod` deployment. Another option would be to allow overriding `branches` config but I can't think of a nice way to achieve that without overriding the whole semantic-release config.

### DIFF
--- a/semantic-release/README.md
+++ b/semantic-release/README.md
@@ -42,7 +42,8 @@ jobs:
 
 Options:
 
-* `DEFAULT_BRANCH` (default: `"main"`): name of the default release branch
+* `DEFAULT_BRANCH` (default: `"main"`): Name of the default release branch
+* `DEFAULT_BRANCH_PRERELEASE` (default: `false`): Whether or not to create a prerelease on the default branch
 * `DRY_RUN` (default: `false`): Runs semantic-release with the `--dry-run` flag to simulate a release but not actually do one
 * `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create GitHub release -- see section below on branch protection for more details
 * `NPM` (default: `false`): Whether or not to release as an NPM package (see "NPM Package Deployment" below for more info)
--- a/semantic-release/action.yml
+++ b/semantic-release/action.yml
@@ -2,8 +2,11 @@ name: Semantic Release
 description: Deploy using semantic-release
 inputs:
   DEFAULT_BRANCH:
-    description: name of the default release branch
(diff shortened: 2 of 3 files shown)


dlockhart:
Interesting. Reading through [semantic-release's docs on pre-releases](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/release-workflow/pre-releases.md#publishing-pre-releases), they seem to recommend using branches specifically named `beta` (and `alpha` if you want two pre-releases at once) for this kind of thing. I'm guessing that would be a major departure in terms of workflow for you though.

One alternate technique that my team has used successfully is to do the dev deployment in the pull request phase. So each commit to a branch actually deploys to a `dev` environment that you can use and test with. Then when the merge to `main` (aka `master`) happens, the production deployment is done. IMO, this is a better mental model since ideally code in `master` is in production or ready to go to production at any point.

AntonBazhal:
> Interesting. Reading through [semantic-release's docs on pre-releases](https://github.com/semantic-release/semantic-release/blob/master/docs/recipes/release-workflow/pre-releases.md#publishing-pre-releases), they seem to recommend using branches specifically named `beta` (and `alpha` if you want two pre-releases at once) for this kind of thing. I'm guessing that would be a major departure in terms of workflow for you though.
> 
> One alternate technique that my team has used successfully is to do the dev deployment in the pull request phase. So each commit to a branch actually deploys to a `dev` environment that you can use and test with. Then when the merge to `main` (aka `master`) happens, the production deployment is done. IMO, this is a better mental model since ideally code in `master` is in production or ready to go to production at any point.

OK, thanks for your feedback. I've spent some time thinking about this and don't believe using `alpha` and `beta` branches makes a lot of sense in our case, but I agree that pre-releasing on the default branch is a temporary workaround for our niche use case and it does not make sense to extend this action. I'm gonna close this PR and I'll put a custom configuration in `package.json` temporarily which should get us going.

dlockhart:
Sounds good Anton, thanks for the good discussion!

PR #68, opened 2022-03-16
Removed `CODEARTIFACT_AUTH_TOKEN`

devpow112:
No longer needed or used.


svanherk (approved):
Nice!  And looks like we can cleanup `NODE_AUTH_TOKEN` if everyone moves over to code artifact one day

PR #70, opened 2022-04-11
Update repo-settings docs for visual-diff

j3parker:


visual-diff/README.md
```diff
@@ -52,22 +52,10 @@ Notes:
...
-
-Specifically, you will need to add the following to `<your_repo>.yaml` in [https://github.com/Brightspace/repo-settings/tree/main/repositories/github](https://github.com/Brightspace/repo-settings/tree/main/repositories/github):
-```
-aws_access:
-  repo:
-    assumable_role_arns:
-      # Dev-UiPlatform
-      - arn:aws:iam::037018655140:role/visual-diff-githubactions-access
-```
+In order to have the `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN` available to you, you will need to add that info to your repo setup in [`repo-settings`](https://github.com/Brightspace/repo-settings).  [See the documentation](https://github.com/Brightspace/repo-settings/blob/main/docs/visual-diff.md).
```
j3parker, line 55: A few changes:

* The process in repo-settings is now a bit simpler (just `visualdiff: true`)
* For a few reasons we want to avoid replicating the docs for repo-settings elsewhere, so I removed the snippet.
* I removed the mention of iam-build-tokens because we don't want people to know about that repo (it is going to become an implementation detail of repo-settings)
* I removed a mention of hub-roles and how you can add other features like deployment. That is true, but e.g. we're streamlining the process for npm as well (see [the docs for that](https://github.com/Brightspace/repo-settings/blob/main/docs/npm.md), which no longer requires you to know about hub roles etc.) and so we want people to get that info straight from repo-settings.
* I removed stuff like the explanation of how it connects to `AWS_ACCESS_KEY_ID`, etc. That stuff is still true, but we're trying to de-emphasize it a bit. I would recommend changing your action to support these coming from environment variables instead (maintaining the input option for backwards compatibility), so that people can set these at the job level instead of individually for each step and not worry about those details.

dlockhart: Awesome!

dlockhart (approved):

PR #75, opened 2022-05-20
adding NPM support to incremental-release

dlockhart:
I think we'll just have to test this out in an actual repo when we convert one over.

I'm not super happy with how different NPM publishing with `semantic-release` is vs. `incremental-release`, which is why their READMEs are _almost_ the same but not similar enough to create a new doc that they both share.

The weirdness comes from the fact that our step to setup the CodeArtifact NPM registry [essentially modifies your local .npmrc file](https://github.com/Brightspace/codeartifact-actions/blob/main/npm/add-registry/action.yml#L20) to use the `CODEARTIFACT_AUTH_TOKEN` as the NPM auth token.

That's a little magical, but it's why `npm publish` "just works". That makes `incremental-release` easy (we just call `npm publish`), but unfortunately `semantic-release` is [doing something similar](https://github.com/semantic-release/npm/blob/7338fc25de82058ea83b1e6e6669707036c2d96a/lib/set-npmrc-auth.js#L41) but it uses the `NPM_TOKEN` that you pass it -- so we need to marshall the CodeArtifact one through there.

One alternative we could try is to just always pass `false` to `semantic-release` and then do the NPM publishing ourselves, which would make the incremental and semantic setups the same. Thoughts?

### DIFF
--- a/incremental-release/README.md
+++ b/incremental-release/README.md
@@ -1,6 +1,6 @@
 # Incremental Release Action
 
-This GitHub action looks for keywords in the latest commit to automatically increment the package version and create a tag.
+This GitHub action looks for keywords in the latest commit to automatically increment the package version, create a GitHub release/tag and optionally publish to NPM.
 
 ## Using the Action
 
@@ -36,6 +36,8 @@ Options:
 * `DEFAULT_INCREMENT` (default: `skip`): If no release keyword is found in the latest commit message, this value will be used to trigger a release. Can be one of: `skip`, `patch`, `minor`, `major`.
 * `DRY_RUN` (default: `false`): Simulates a release but does not actually do one
 * `GITHUB_TOKEN`: Token to use to update version in 'package.json' and create the tag -- see section below on branch protection for more details
+* `NPM` (default: `false`): Whether or not to release as an NPM package (see "NPM Package Deployment" below for more info)
+* `NPM_TOKEN` (optional if `NPM` is `false` or publishing to CodeArtifact): Token to publish to NPM (see "NPM Package Deployment" below for more info)
 
 Outputs:
 * `VERSION`: will contain the new version number if a release occurred, empty otherwise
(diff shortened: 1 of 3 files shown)


incremental-release/README.md
```diff
@@ -49,6 +51,47 @@ The release step will fail to write to `package.json` if you have branch protect
...
+
+### NPM
+
+Setup Node with the `registry-url` option:
+
+```yml
+- name: Setup Node
+  uses: Brightspace/third-party-actions@actions/setup-node
+    with:
+     registry-url: 'https://registry.npmjs.org'
```
dlockhart, line 70: I could have done the same thing `semantic-release` does and much around in the local `.npmrc` file to add the registry, but this seemed more like how GitHub Actions intended this to be done:
https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

svanherk: So I was gonna say, if we were ok with limiting the actions, we could get the two more similar by assuming in `semantic-release` that if you have `npm: true` but don't pass in an `npm_token`, we should use the codeartifact one and pass that to the `semantic-release` action.  But that doesn't solve the difference above anyways

svanherk: Not sure why that posted twice... weird

dlockhart: Honestly I wish `semantic-release` was less magical and then they could both work the same way.

svanherk: Would it be if we turned off the `npm` plugin and did it ourselves?  Or is that what also updates `package.json`?

dlockhart: Correct, it's what also updates `package.json` (which we do ourselves in incremental release). I did consider that, but that change would involve going into every repo using semantic-release + NPM and add the `registry-url: 'https://registry.npmjs.org'` line to `setup-node`.

svanherk: Right.  I've been thinking of how we could kill `brightspace-bot` if we handled the update to `package.json` ourselves, using some fancy status check work Jacob added to the `lms` repo.  So maybe one day we could tie those changes together in a new version of the action

svanherk (approved):
I'm good with the slight differences to keep things a bit cleaner

PR #78, opened 2022-06-10
Automatically release with the LMS

dlockhart:
Adding this feature to the `semantic-release` action!

### DIFF
--- a/semantic-release/action.yml
+++ b/semantic-release/action.yml
@@ -10,11 +10,16 @@ inputs:
   GITHUB_TOKEN:
     description: Token to use to update version in 'package.json' and create GitHub release
     required: true
+  MINOR_RELEASE_WITH_LMS:
+    description: Automatically perform a minor release whenever the LMS release changes (requires RALLY_API_KEY)
+    default: false
   NPM:
     description: Whether or not to release as an NPM package
     default: false
   NPM_TOKEN:
     description: Token to publish to NPM
+  RALLY_API_KEY:
+    description: Key to access the Rally API, required for MINOR_RELEASE_WITH_LMS
 outputs:
   VERSION:
     description: Version of the new release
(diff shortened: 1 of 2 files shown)


semantic-release/action.yml
```diff
@@ -24,17 +29,37 @@ runs:
   steps:
     - name: Installing semantic-release
       run: |
-        if [ ${{ !contains(github.event.head_commit.message, 'skip ci') }} == true ]; then
```
dlockhart, line 27: I confirmed that these checks aren't needed anymore, so was able to clean up a few of the steps in here.

semantic-release/action.yml
```diff
@@ -24,17 +29,37 @@ runs:
...
+      run: echo "::set-output name=value::$(cat .lmsrelease)"
+      shell: bash
+    - name: Get Active LMS Release
+      if: ${{ inputs.MINOR_RELEASE_WITH_LMS == 'true' }}
+      id: get-lms-version
+      uses: Brightspace/lms-version-actions/get-lms-release@main
+      with:
+        RALLY_API_KEY: ${{ inputs.RALLY_API_KEY }}
+    - name: Update Repo's Last LMS Release
+      if: ${{ inputs.MINOR_RELEASE_WITH_LMS == 'true' && inputs.DRY_RUN == 'false' && steps.repo-lms-version.outputs.VALUE != steps.get-lms-version.outputs.LMS_VERSION }}
```
svanherk, line 47: So seems like `DRY_RUN` won't tell you if it would have done a minor release instead of the one you specified, which probably isn't a big deal and a low likelihood of ever happening in real life

semantic-release/action.yml
```diff
@@ -24,17 +29,37 @@ runs:
...
+      run: |
+        echo "Updating last LMS release..."
+        echo "${{ steps.get-lms-version.outputs.LMS_VERSION }}" > .lmsrelease
+        git config user.name github-actions[bot]
+        git config user.email 41898282+github-actions[bot]@users.noreply.github.com
+        git add .lmsrelease
+        git commit -m "feat: updating tracked LMS release version [skip ci]"
+        git push --follow-tags --repo=https://$GITHUB_TOKEN@github.com/${{ github.repository }}
```
svanherk, line 55: And just so I'm understanding - the next merge after a release will always do a minor bump, even if it was going to do no bump at all (like a docs change or dependabot PR merge)

dlockhart: Yes, correct -- any merge to `main`. This basically interjects a commit with `feat: ` before semantic-release has a look, which will always cause the minor bump.

I think that's what we want, since a Dependabot change is still something we'd want to isolate from old releases.

semantic-release/action.yml
svanherk: Yeah I don't think its worth it at the moment

PR #81, opened 2022-06-22
visual-diff: add retries

margaree:
Tested in this PR https://github.com/BrightspaceUI/core/pull/2570

Something to note is that if we're in a situation where a lot of tests are failing and retrying (e.g., puppeteer version bump) this could result in a timeout. In this situation we could bump `timeout-minutes` temporarily in the visual-diff workflow.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -16,6 +16,9 @@ inputs:
   GITHUB_TOKEN:
     description: Token to use to open the goldens PR
     required: true
+  RETRIES:
+    description: Number of times to retry a failed test
+    default: 2
   TEST_PATH:
     description: Path passed into the mocha call defining the locations and name structure of the tests
     default: './{,!(node_modules)/**}/*.visual-diff.{js,mjs}'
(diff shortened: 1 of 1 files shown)


svanherk:
> Something to note is that if we're in a situation where a lot of tests are failing and retrying (e.g., puppeteer version bump) this could result in a timeout. In this situation we could bump timeout-minutes temporarily in the visual-diff workflow.

Or since you made this an input, we could also just set retries to 1 temporarily 🎉

margaree:
> > Something to note is that if we're in a situation where a lot of tests are failing and retrying (e.g., puppeteer version bump) this could result in a timeout. In this situation we could bump timeout-minutes temporarily in the visual-diff workflow.
> 
> Or since you made this an input, we could also just set retries to 1 temporarily 🎉

Yup absolutely!

dlockhart:
Can you update the README? Typically we list out all the options and what they do.

dlockhart (approved):
Yay!

PR #83, opened 2022-06-29
Add get-maintenance-version action

eKoopmans:
This is a generic action that can be reused for checking whether the current branch is a "maintenance branch" and getting the matching maintenance version.

It's a prerequisite for getting translations merged into BSI (and the LMS) for 2022.7. This is the plan:

1. Add this action
2. Update the `semantic-release` action to skip the `MINOR_RELEASE_WITH_LMS` steps when it's a maintenance branch
3. On BrightspaceUI/core, delete and recreate branch `2.25.x` - this should retrigger the action, with this change, causing a patch version bump and release

Rally: https://rally1.rallydev.com/#/?detail=/userstory/641415440391&fdp=true

### DIFF
--- a/get-maintenance-version/README.md
+++ b/get-maintenance-version/README.md
@@ -13,7 +13,7 @@ Include this action at any point before its data is required, and retrieve the d
 - name: Some other action
   run: |
     echo "Maintenance version: ${{ steps.get-maintenance-version.outputs.MAINTENANCE_VERSION }}"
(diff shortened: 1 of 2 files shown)


dlockhart (approved):
Cool! I could go either way as to whether this action belongs here in BrightspaceUI or [with the other lms-version actions](https://github.com/Brightspace/lms-version-actions). This isn't really LMS-specific, so I'm good with it being here. 🤷

dlockhart:
> They're both public repos, so I don't think so.

That's my understanding as well!

PR #84, opened 2022-06-30
Skip minor release on maintenance branches

eKoopmans:
This is a follow-up to #83.

Adding the `IS_MAINTENACE_VERSION` check individually to each step isn't the cleanest, but it follows what's being done already with `MINOR_RELEASE_WITH_LMS`. The alternative would be moving that chunk of steps to its own action, and controlling whether to run the whole action with those two checks.

Rally: https://rally1.rallydev.com/#/?detail=/userstory/641415440391&fdp=true


dlockhart (approved):
🙌

PR #76, opened 2022-05-25
Revert "adding NPM support to incremental-release (#75)"

dlockhart:
This reverts commit 4f846c0877d0a3eb642ea8a9b441224a0f5031e4.

It was causing issues as I forgot the `run: |` section, but we're also now not sure we even need this so I'm just removing it entirely for now.

PR #91, opened 2022-09-12
feat: Add npm publish with tags to incremental-release

eKoopmans:
This PR:

- re-adds the optional npm publish step that was originally added (then removed) by @dlockhart in PRs:
  - https://github.com/BrightspaceUI/actions/pull/75
  - https://github.com/BrightspaceUI/actions/pull/76
- adds logic for tagging any npm release coming from a "maintenance branch", to match what's now being done in `match-lms-version`:
  - https://github.com/Brightspace/lms-version-actions/pull/6
- updates `get-maintenance-version` to recognize maintenance branches that don't end with `.x`
  - e.g. `release/3.19`, which is the relevant branch that needs fixing currently

Rally: https://rally1.rallydev.com/#/?detail=/userstory/646357845471&fdp=true

### DIFF
--- a/get-maintenance-version/action.yml
+++ b/get-maintenance-version/action.yml
@@ -12,7 +12,7 @@ runs:
   steps:
     - id: main
       run: |
-        MAINTENANCE_VERSION=$(echo "${GITHUB_REF##*/}" | grep -Eo "^[0-9]+\.([0-9]+\.)?x$" || true)
+        MAINTENANCE_VERSION=$(echo "${GITHUB_REF##*/}" | grep -Eo "^[0-9]+(\.[0-9]+)?(\.x)?$" || true)
         echo "Maintenance version determined from branch name: $MAINTENANCE_VERSION"
         echo "::set-output name=MAINTENANCE_VERSION::$MAINTENANCE_VERSION"
       shell: bash
--- a/incremental-release/README.md
+++ b/incremental-release/README.md
@@ -1,6 +1,6 @@
 # Incremental Release Action
 
-This GitHub action looks for keywords in the latest commit to automatically increment the package version and create a tag.
+This GitHub action looks for keywords in the latest commit to automatically increment the package version, create a GitHub release/tag, and optionally publish to NPM.
 
 ## Using the Action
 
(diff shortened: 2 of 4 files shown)


incremental-release/README.md
```diff
@@ -49,6 +51,61 @@ The release step will fail to write to `package.json` if you have branch protect
...
+    auth-token: ${{env.CODEARTIFACT_AUTH_TOKEN}}
+```
+
+### NPM
+
+Setup Node:
+
+```yml
+- name: Setup Node
+  uses: Brightspace/third-party-actions@actions/setup-node
```
svanherk, line 83: Did you find that the `https://registry.npmjs.org` stuff from the old PR (https://github.com/BrightspaceUI/actions/pull/75/files#diff-df91ce91fa6ed4c9d6d931dc5cdb55c121dbc990bb0d81a9703fb0587fc2f979R64) wasn't actually needed?

get-maintenance-version/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
   steps:
     - id: main
       run: |
-        MAINTENANCE_VERSION=$(echo "${GITHUB_REF##*/}" | grep -Eo "^[0-9]+\.([0-9]+\.)?x$" || true)
+        MAINTENANCE_VERSION=$(echo "${GITHUB_REF##*/}" | grep -Eo "^[0-9]+(\.[0-9]+)?(\.x)?$" || true)
```
svanherk, line 15: This piece will be interesting to keep an eye out for as places are updated, since unfortunately we don't enforce any naming standard for the maintenance branches like we do with `semantic-release` and `lms-version-release`.  I guess the "worst case" behavior if someone uses a unique name is that they're in the same place as before this feature was added (a certfix version being tagged as latest)

svanherk (approved):

incremental-release/README.md
eKoopmans: Yep, see these comments from @devpow112 :
https://github.com/Brightspace/lms-version-actions/pull/6#discussion_r906364903

get-maintenance-version/action.yml
eKoopmans: Yeah... We could more generically just look for "not main or master", and tag the release with some generic "maintenance" tag, because the specifics of the tag aren't important for us 🤷. Like you say, something to watch.

PR #94, opened 2022-10-14
switch actions to use new output format

dlockhart:


devpow112 (approved):

incremental-release/action.yml
```diff
@@ -54,7 +54,7 @@ runs:
         else
           echo "Did not find commit message flag, using DEFAULT_INCREMENT (${{ inputs.DEFAULT_INCREMENT }})"
         fi
-        echo "::set-output name=increment::$(echo "$TYPE")"
+        echo "increment=$(echo "$TYPE")" >> $GITHUB_OUTPUT
```
eKoopmans, line 57: Same comment about re-echoing - I think now's a good time to remove these. Can't think of any reason why to keep it (except for the `$(cat .lmsrelease)` one, which is doing something different).

semantic-release/action.yml
```diff
@@ -74,7 +74,7 @@ runs:
         GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
         NPM_TOKEN: ${{ inputs.NPM_TOKEN }}
       run: |
-        echo "::set-output name=version::"
+        echo "version::=" >> $GITHUB_OUTPUT
```
eKoopmans, line 77: Missed the `::` removal here.

update-package-lock/action.yml
```diff
@@ -52,7 +52,7 @@ runs:
         id: check-for-changes
         run: |
           echo -e "\n\e[34mChecking For Changes"
-          git diff --exit-code package-lock.json && echo "::set-output name=changes-exist::$(echo false)" || echo "::set-output name=changes-exist::true"
+          git diff --exit-code package-lock.json && echo "changes-exist=$(echo false)" >> $GITHUB_OUTPUT || echo "changes-exist=true" >> $GITHUB_OUTPUT
```
devpow112, line 55: Same double echoing of stuff here

devpow112:
There's a bunch of `echo $(echo $VARIABLE)` type of stuff. All of it was pre-existing. Maybe can be cleaned up in another pass.

incremental-release/action.yml
dlockhart: Ok I've cleaned these up too!

eKoopmans (approved):

devpow112 (approved):

PR #99, opened 2022-11-18
Clarify brightspace-bot instructions

eKoopmans:
Just a small change here to clarify how to set up brightspace-bot and branch protection.

### DIFF
--- a/docs/branch-protection.md
+++ b/docs/branch-protection.md
@@ -14,9 +14,9 @@ To set up this bypass:
 
 2. Configure access to the `brightspace-bot`
 
-    This is also a temporary step, but make sure that the `brightspace-bot` GitHub account has Admin access to your repo.
+    This is also a temporary step, but make sure that the `brightspace-bot` GitHub account has Admin access to your repo by going to Settings -> Collaborators and teams -> Add people.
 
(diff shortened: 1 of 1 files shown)


oelhanafey (commented):
The issue I see with adding specific navigation instructions like this is that changes to the GitHub UI could make these instructions obsolete and confusing, when the user could search through GitHub documentation to figure out how to do this which would (hopefully) be kept up to date.

eKoopmans:
> changes to the GitHub UI could make these instructions obsolete and confusing

Hmm... My take is that they'd be no less confusing than there being no instructions.

> the user could search through GitHub documentation to figure out how to do this

I added these notes from my experience, where it wasn't obvious what to search for in the GitHub docs and there wasn't a clear doc with the answers I was looking for.

I'm hoping that with these changes, the process is ideally one-stop just looking at this doc - then if the GitHub UI has changed, it'll take a bit more hunting, but these instructions will give a starting point. And then if the user's feeling compelled, they can update this doc 😄.

oelhanafey (approved):
Okay makes sense! Yeah I suppose if the UI did change, then the user would realize that quickly, search elsewhere and hopefully update the doc.

PR #100, opened 2022-12-02
Add removed dependencies to dependency changes table 

imcdonald-d2l:
Add removed dependencies to the dependency changes table on the PR created by the action. Tested off dummy data locally - this should work just fine but has not been tested in a "production" environment.


svanherk (approved):
🎉🎉🎉 Thanks Ian!

PR #101, opened 2023-01-11
Update docs to remove D2L_GITHUB_TOKEN

svanherk:
We want to remove all unnecessary usages of `D2L_GITHUB_TOKEN`.  This action doesn't require it.


dlockhart (approved):
Nice

PR #102, opened 2023-01-12
Adjust AUTO_MERGE_METHOD commit message description

devpow112:
Now that repositories let you set what the default message for squash merges are update docs to point to the documentation on that and simplify description.


dlockhart (approved):
Thanks!

PR #106, opened 2023-03-27
US150506 - Increase semantic-release version

svanherk:
Moving from version `17` to `19` means that everything below `Node 14.17` and `Node 15` are now unsupported. We can't move to version `20` or `21` yet because they drop support for everything under `Node 18`, which would break a large number of our repos.

### DIFF
--- a/semantic-release/README.md
+++ b/semantic-release/README.md
@@ -51,6 +51,7 @@ Outputs:
 * `VERSION`: will contain the new version number if a release occurred, empty otherwise
 
 Notes:
+* This action currently requires `Node` version `16` or higher (or version `^14.17`).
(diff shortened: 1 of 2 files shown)


semantic-release/README.md
```diff
@@ -51,6 +51,7 @@ Outputs:
 * `VERSION`: will contain the new version number if a release occurred, empty otherwise
 
 Notes:
+* This action currently requires `Node` version `16` or higher (or version `^14.17`).
```
svanherk, line 54: I don't love how I worded this - suggestions?  I basically want to say "Get to Node 16, 15 is unsupported.  14 can work if it must."

svanherk:
> Is this going to immediately break any CI that's using this action with an earlier version of node?

Yes, but I did an audit and found only one place using node 12 still.  It's been updated to Node 18

PR #107, opened 2023-03-29
Fix `npm ls` related issues

devpow112:
The `npm ls` command will return a non-zero error code if there are issues like unmet dependencies or other things it deems not good. This will fail the current action execution even though the `npm install` command was successful. This makes those commands always succeed and also does some additional checking on the results of the generated JSON.

### DIFF
--- a/update-package-lock/action.yml
+++ b/update-package-lock/action.yml
@@ -32,7 +32,7 @@ runs:
         run: |
           echo -e "\e[34mSaving Initial Dependencies"
           npm ci
-          npm ls --json --package-lock-only --all > "$RUNNER_TEMP/dependencies-before.json"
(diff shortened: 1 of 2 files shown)


devpow112:
NM that causes badness on the next `npm ci`. Guess it's better to have it error early instead of opening the PR and failing. FYI something is up with `puppeteer`, it seems to be causing the issue.

https://github.com/Brightspace/d2l-rubric/actions/runs/4554780071/jobs/8033118950?pr=2609

devpow112:
Running a second `npm i` after running the update seems to fix the issue. For some reason deleting `node_modules` and running `npm i` make an invalid setup for `typescript`. Re-running `npm i` fixes the dependency. No idea why but it works.

PR #112, opened 2023-06-14
Add CODEOWNERS clarification for update-package-lock

eKoopmans:
I remember this step of setting up `update-package-lock` being a bit confusing, and just saw a question about it over on https://github.com/Brightspace/Mobile-API/pull/1512, so I added some clarification.


svanherk (approved):
Much better, thanks @eKoopmans!

PR #113, opened 2023-07-18
US152262 - First step of new vdiff action

svanherk:
I'll add a step at a time, using the old action as a base.  We no longer need to do the [first step](https://github.com/BrightspaceUI/actions/blob/main/visual-diff/action.yml#L31-L40) of the old action that installed dependencies. Then the next step (added here) is basically a copy from the existing action:
- https://github.com/BrightspaceUI/actions/blob/main/visual-diff/action.yml#L42-L49
- https://github.com/BrightspaceUI/actions/blob/main/visual-diff/cleanup-branches.js

### DIFF
--- a/vdiff/README.md
+++ b/vdiff/README.md
@@ -1,0 +1,3 @@
+# vdiff Action
+
+TO DO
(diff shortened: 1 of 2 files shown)


vdiff/README.md
```diff
@@ -0,0 +1,3 @@
+# vdiff Action
+
+TO DO
```
svanherk, line 3: Will fill out the README at the end once the action is stable.

bearfriend (approved):

PR #114, opened 2023-07-25
US152262 - Add steps to copy goldens to .vdiff, then copy them back after tests run

svanherk:
These steps don't exist in the current action. We need to copy all the goldens from their stored locations in `golden` folders to the proper location in `.vdiff`.  Then the tests will run (added in the next PR).  Then we'll move the passed folders back as is, and copy the failed tests in.

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -64,3 +64,69 @@ runs:
       env:
         FORCE_COLOR: 3
         PREFIX: ${{ inputs.VDIFF_BRANCH_PREFIX }}
+
+    - name: Copy Current Goldens
+      run: |
+        echo -e "\e[34mCopying Current Goldens to .vdiff"
+        echo "Clearing .vdiff directory"
+        rm -rf .vdiff
+
+        echo "Moving current goldens to .vdiff directory"
+        find . -name golden -type d | while read GOLDEN_DIR; do
+          TEST_PATH=`dirname "${GOLDEN_DIR}"`
+          TEST_PATH=${TEST_PATH:2}
+          for TEST in `ls "${GOLDEN_DIR}"`; do
+            mkdir -p "./.vdiff/${TEST_PATH}/${TEST}/golden"
+            mv "${GOLDEN_DIR}/${TEST}/"* "./.vdiff/${TEST_PATH}/${TEST}/golden"
+          done
+        done
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
+    - name: Run vdiff Tests and Upload Report
+      run: |
+        echo -e "\e[34mRunning vdiff Tests"
+        echo To Do
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
+    - name: Copy New Goldens
+      run: |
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -64,3 +64,69 @@ runs:
...
+          for TEST in `ls "${GOLDEN_DIR}"`; do
+            mkdir -p "./.vdiff/${TEST_PATH}/${TEST}/golden"
+            mv "${GOLDEN_DIR}/${TEST}/"* "./.vdiff/${TEST_PATH}/${TEST}/golden"
+          done
+        done
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
+    - name: Run vdiff Tests and Upload Report
```
svanherk, line 87: This will eventually be 3 steps - running the tests, assuming the AWs role and uploading the report.  But I thought it'd be better to put these steps in together for context, and add this as a placeholder.

vdiff/action.yml
```diff
@@ -64,3 +64,69 @@ runs:
       env:
         FORCE_COLOR: 3
         PREFIX: ${{ inputs.VDIFF_BRANCH_PREFIX }}
+
+    - name: Copy Current Goldens
```
dlockhart, line 68: Should this be "Move Current Goldens"?

vdiff/action.yml
```diff
@@ -64,3 +64,69 @@ runs:
...
+      run: |
+        echo -e "\e[34mCopying Current Goldens to .vdiff"
+        echo "Clearing .vdiff directory"
+        rm -rf .vdiff
+
+        echo "Moving current goldens to .vdiff directory"
+        find . -name golden -type d | while read GOLDEN_DIR; do
```
dlockhart, line 75: If some poor person creates a directory called `golden` this will cause them a world of pain... haha but that's probably fine.

vdiff/action.yml
```diff
@@ -64,3 +64,69 @@ runs:
...
+
+    - name: Run vdiff Tests and Upload Report
+      run: |
+        echo -e "\e[34mRunning vdiff Tests"
+        echo To Do
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
+    - name: Copy New Goldens
```
dlockhart, line 95: Technically this is also "moving" not "copying". 🤷

dlockhart (approved):

vdiff/action.yml
svanherk: Definitely lol, but yeah not sure there's much we can do about it.  We have to pick _something_ to call it.  We could make it safer by removing flexibility (or example, it'll only check inside folders called "test" or something).  Although it should hopefully be pretty obvious what's going when you get the disaster of a PR

dlockhart: Yeah I'm not actually worried haha. If it ends up being a problem, maybe we only move things that contain a `*.vdiff.js`.

vdiff/action.yml
svanherk: Updated - wrote these headings before I realized moving was better, since it naturally cleans up orphans that way

dlockhart (approved):

PR #115, opened 2023-07-25
US152262 - Add steps to run the vdiff tests, assume the AWS role and upload the report

svanherk:
The [current "run tests" step](https://github.com/BrightspaceUI/actions/blob/edfa81de66bde59d9bfd397f2bae15c2ba3f8de1/visual-diff/action.yml#L53-L66) handles running the mocha command directly, taking in each option as an input. The AWS steps are new, because previously the `visual-diff` library itself did the uploading.

Lots of questions about these pieces - will add comments inline.

Related: https://github.com/Brightspace/d2l.dev/pull/4574

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -102,6 +102,21 @@ runs:
         FORCE_COLOR: 3
       shell: bash
 
+    - name: Prepare Report
+      id: prepare-report
+      if: ${{ inputs.AWS_ACCESS_KEY_ID }}
+      continue-on-error: true
+      run: |
+        echo -e "\e[34mPreparing vdiff Report"
+        mv "./.vdiff/.report/" "./.vdiff/report/"
+
+        DATE_TIME=$(date '+%Y-%m-%d_%H-%M-%S');
+        UPLOAD_PATH="${{ github.repository }}/${{ github.sha }}/${DATE_TIME}"
+        echo "upload-path=${UPLOAD_PATH}" >> ${GITHUB_OUTPUT}
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
     - name: Assume vdiff Role for Uploading Report
       if: ${{ inputs.AWS_ACCESS_KEY_ID }}
       continue-on-error: true
@@ -118,19 +133,11 @@ runs:
       id: upload-report
       if: ${{ inputs.AWS_ACCESS_KEY_ID }}
       continue-on-error: true
-      run: |
-        echo -e "\e[34mUpload vdiff Report"
-        mv "./.vdiff/.report/" "./.vdiff/report/"
-
-        DATE_TIME=$(date '+%Y-%m-%d_%H-%M-%S');
-        UPLOAD_PATH="${{ github.repository }}/${DATE_TIME}-${{ github.sha }}"
-        aws s3 sync --delete ./.vdiff/ s3://visual-diff.d2l.dev/reports/${UPLOAD_PATH}/
-        echo "report-path=https://vdiff.d2l.dev/${UPLOAD_PATH}/report/" >> ${GITHUB_OUTPUT}
-      env:
-        # cred variables set in the "Assume role" step
-        AWS_DEFAULT_REGION: ca-central-1
-        FORCE_COLOR: 3
-      shell: bash
+      uses: BrightspaceUI/actions/publish-to-s3@main
+      with:
+        BUCKET_PATH: s3://visual-diff.d2l.dev/reports/${{ steps.prepare-report.outputs.upload-path }}/
+        CACHE_DEFAULT: "--cache-control max-age=120"
+        PUBLISH_DIRECTORY: ./.vdiff/
 
     - name: Move New Goldens
       run: |
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -1,6 +1,15 @@
 name: vdiff
 description: Run your vdiff tests, upload a report and open a PR with the new goldens as necessary
 inputs:
+  AWS_ACCESS_KEY_ID:
+    description: Access key id for the role that will assume the vdiff role
+    required: true
+  AWS_SECRET_ACCESS_KEY:
+    description: Access key secret for the role that will assume the vdiff role
+    required: true
+  AWS_SESSION_TOKEN:
+    description: Session token for the role that will assume the vdiff role
+    required: true
   GITHUB_TOKEN:
     description: Token used to cleanup branches and open the goldens PR
     required: true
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -84,14 +93,46 @@ runs:
         FORCE_COLOR: 3
       shell: bash
 
-    - name: Run vdiff Tests and Upload Report
+    - name: Run vdiff Tests
+      id: test-run
       run: |
         echo -e "\e[34mRunning vdiff Tests"
-        echo To Do
+        npm run test:vdiff && echo "passed=true" >> ${GITHUB_OUTPUT} || echo "passed=false" >> ${GITHUB_OUTPUT}
```
svanherk, line 100: So I wasn't sure what we wanted to do here.  This works nicely, but there's pros and cons to each method:

1. Let them define `test:vdiff`, and they can add whatever flags they want.  Timeout, browsers, etc.
2. We run `d2l-test-runner --group vdiff` and expose all the flags we want them to be allowed to override as inputs
3. We take their whole command as an input - this was my least favourite, but maybe we let them pass all the flags as one input string and still control the main command?

vdiff/action.yml
```diff
@@ -84,14 +93,46 @@ runs:
...
+      env:
+        FORCE_COLOR: 3
+      shell: bash
+
+    - name: Assume vdiff Role for Uploading Report
```
svanherk, line 105: We skip the next two steps if they don't pass in `AWS_ACCESS_KEY_ID`.  If they do but it's expired or incorrect, we `continue-on-error`.  In both cases, we'll let them know we couldn't upload the report and point them to docs to set things up.

vdiff/action.yml
```diff
@@ -84,14 +93,46 @@ runs:
...
+      id: upload-report
+      if: ${{ inputs.AWS_ACCESS_KEY_ID }}
+      continue-on-error: true
+      run: |
+        echo -e "\e[34mUpload vdiff Report"
+        mv "./.vdiff/.report/" "./.vdiff/report/"
+
+        DATE_TIME=$(date '+%Y-%m-%d_%H-%M-%S');
+        UPLOAD_PATH="${{ github.repository }}/${DATE_TIME}-${{ github.sha }}"
+        aws s3 sync --delete ./.vdiff/ s3://visual-diff.d2l.dev/reports/${UPLOAD_PATH}/
```
svanherk, line 127: I'll have to test this after merging.  Also, was there any kind of compression or caching we want?

vdiff/action.yml
```diff
@@ -84,11 +93,42 @@ runs:
...
+      uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+      with:
+        aws-access-key-id: ${{ inputs.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ inputs.AWS_SECRET_ACCESS_KEY }}
+        aws-session-token: ${{ inputs.AWS_SESSION_TOKEN }}
+        role-to-assume: "arn:aws:iam::037018655140:role/visual-diff-githubactions-access"
+        role-duration-seconds: 3600
+        aws-region: ca-central-1
+
+    - name: Upload Report
```
svanherk, line 117: We rename the `.report` folder to `report`, then upload the whole `.vdiff` folder. It'll look like this in S3:
```
visual-diff.d2l.dev/reports/Brightspace/copy-role/2023-07-25_21-47-34-3f72c233dfe1a0ed1a744d9dddd2d809625b92d8/*
```
And that would be accessed at:
```
https://vdiff.d2l.dev/Brightspace/copy-role/2023-07-25_21-47-34-3f72c233dfe1a0ed1a744d9dddd2d809625b92d8/report/
```

vdiff/action.yml
dlockhart: It's a bit weird IMO assuming that their `package.json` has a script called `test:vdiff`, especially since we don't have a great track record with being consistent about the names of those common scripts.

I think I personally like exposing the command as an input but defaulting it to `npm run d2l-test-runner --group vdiff`. That way if they want to override that and pass different things they can, or they can just override it to `npm run test:vdiff` if they've already defined all the things they want there.

vdiff/action.yml
dlockhart: Thoughts about inverting the SHA and the timestamp and putting a `/` between them? So:

> https://vdiff.d2l.dev/Brightspace/copy-role/3f72c233dfe1a0ed1a744d9dddd2d809625b92d8/2023-07-25-21-47-34/report/

If we ever wanted to turn on directory indexes for the S3 buckets, that'd let you see a list of all the reports for a particular SHA.

vdiff/action.yml
bearfriend: Letting them run any command seems weird, but maybe we can do some basic validation on it. Otherwise, I don't hate option 2.

vdiff/action.yml
svanherk: Do you remember if there was a reason we didn't initially? It _seems_ good to go...

dlockhart: Erik's comment from the original PR that added it:
> I've designed both actions to be theoretically reusable in the future, but keeping them internal for now. No readmes for them at the moment as they're just internal.

vdiff/action.yml
svanherk: My thinking was I'd rather have them in order based on time in the bucket, but we'd probably be able to get the same thing by sorting by "Last Modified" since we never overwrite.

Also, worth pointing out that this SHA is not as discoverable for PRs - it's the last merge commit on the `refs/pull/<pr_num>/merge` branch.  That's what we were using before in old `visual-diff`, but if we want these SHAs to be more useful we could instead use the commit of the branch that triggered them

vdiff/action.yml
svanherk: Done 👍

vdiff/action.yml
svanherk: Ok, I've switched the sha and the date and updated to a more intuitive sha to use depending on whether the vdiffs are against a branch or a PR

vdiff/action.yml
```diff
@@ -1,19 +1,57 @@
...
+  test-timeout:
+    description: Timeout threshold for vdiff tests (in ms)
+    default: 40000
+  vdiff-branch-prefix:
     description: Prefix for vdiff branches
     default: 'ghworkflow/vdiff'
 runs:
   using: composite
   steps:
+    - name: Get Run Info
```
svanherk, line 37: There will be [more info here soon](https://github.com/BrightspaceUI/actions/blob/main/visual-diff/action.yml#L71-L79), and I use the original-sha for a bunch of other things anyways.  Adding it now to use for the report upload path

vdiff/action.yml
dlockhart: I like it!

svanherk: Yeah like theoretically, we can change the name of `d2l-test-runner`, or switch back to a single browsers flag, or whatever and they don't need to change their action.  But they would need to get the latest version of `testing`, so... 🤷

vdiff/action.yml
```diff
@@ -1,19 +1,57 @@
...
+  github-token:
     description: Token used to cleanup branches and open the goldens PR
     required: true
-  VDIFF_BRANCH_PREFIX:
+  test-chrome:
+    description: Run vdiff tests in Chromium
+    default: true
+  test-files:
+    description: Pattern of vdiff tests to run
+    default: './test/**/*.vdiff.js'
```
bearfriend, line 21: Is this going to override the `pattern` in the config file?

svanherk: Looks like yes, so I guess maybe I don't default any of these? And I only pass flags if you set them to something, saying you want your config overridden? 🤔

Realizing I also don't let you override the config, cause I figure no one will ever really need to to do that.  But I could add it

bearfriend: I was thinking about `--config`. I could be useful in a monorepo?

I think the browser stuff seems fine since it's constructing the options only when they're set, but yeah `--files` can't have a default, so you'll have to do that similarly.

svanherk: Ok, updated to default to false/nothing and not pass the flag unless they tell us to override their config/the runner's defaults.

We could just have one blanket `flags` input, but I think I still like how this would allow us to change things under the hood

dlockhart: If we ever want to rename/remove an input, it'll also make it easier for us to search for usages.

dlockhart (approved):

bearfriend (approved):

PR #116, opened 2023-08-02
US155896 - Add publish-to-s3 action from core

svanherk:
This is [lifted from `core`](https://github.com/BrightspaceUI/core/blob/03dfe28bf436cf26e329c38066dee2e1b692e594/.github/actions/publish-to-s3/action.yml#L1). The only changes I made were moving the `env` variables after `run` to match the layout of all the other actions in this repo, and adding the `README`. 

Originally I had an output for whether the publish was successful or not, but I think I can get that using a mix of `continue-on-error` and `failure()`.  If not I'll add it back here.

I'll put up a PR in `core` to switch to this, and update the `vdiff` PR to use it.


dlockhart (approved):
Nice! Kudos to Erik for making this so transferable in the first place. 👏

PR #118, opened 2023-08-08
Allow setting working directory for auto update action

devpow112:
~No idea if this will work yet just want to try it out.~

This works as expected. Allows you to specify a sub directory to work within. See https://github.com/Brightspace/test-reporting.d2l.dev/pull/119 for example.

### DIFF
--- a/update-package-lock/action.yml
+++ b/update-package-lock/action.yml
@@ -25,21 +25,27 @@ inputs:
   PR_TITLE:
     description: Title for the opened pull request
     default: 'Updating package-lock.json'
+  WORK_DIRECTORY:
+    description: Path where you want to execute the update action
+    default: '.'
 runs:
   using: composite
   steps:
       - name: Save Initial Dependencies
+        working-directory: ${{ inputs.WORK_DIRECTORY }}
         run: |
           echo -e "\e[34mSaving Initial Dependencies"
           npm ci
           npm ls --json --package-lock-only --all > "$RUNNER_TEMP/dependencies-before.json"
         shell: bash
       - name: Install Dependencies
+        working-directory: ${{ inputs.WORK_DIRECTORY }}
         run: |
           echo -e "\e[34mInstalling Dependencies"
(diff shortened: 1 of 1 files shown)


update-package-lock/action.yml
```diff
@@ -25,28 +25,35 @@ inputs:
   PR_TITLE:
     description: Title for the opened pull request
     default: 'Updating package-lock.json'
+  WORKING_DIRECTORY:
+    description: Path where you want to execute the update action
+    default: '.'
```
devpow112, line 30: Just default to current working directory

svanherk:
I think we might be able to just set this once, not every step: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-set-the-default-shell-and-working-directory

svanherk:
Ah nvm, looks like we can't use `inputs` in there 😞

devpow112:
> ~I think we might be able to just set this once, not every step: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-set-the-default-shell-and-working-directory~
> 
> Ah nvm, looks like we can't use `inputs` in there 😞

Yeah for composite actions you have to set it for every step sadly.

update-package-lock/action.yml
```diff
@@ -78,14 +87,15 @@ runs:
...
-          
+
           echo -e "\e[34mSaving New Dependencies"
           npm ls --json --package-lock-only --all > "$RUNNER_TEMP/dependencies-after.json"
         env:
           GITHUB_TOKEN: ${{ inputs.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{ inputs.NODE_AUTH_TOKEN }}
         shell: bash
       - name: Create PR (if necessary)
+        working-directory: ${{ inputs.WORKING_DIRECTORY }}
```
svanherk, line 98: I figured one day we'd move to using `github-script` for this step rather than the `js` file, which might make things a bit more complicated if `working-directory` is actually needed for this step.  But I'm fine with figuring that out at that time

svanherk (approved):

PR #119, opened 2023-08-08
feat: Add comment-on-pr action from core

eKoopmans:
I'm working on making "PR Previews" into a more reusable pattern that all teams can easily implement. Part of that is moving the required actions out of core to somewhere more central.

This PR:

- [copies `comment-on-pr` from core](https://github.com/BrightspaceUI/core/blob/main/.github/actions/comment-on-pr/action.yml)
- updates inputs/outputs to lowercase
- adds a readme

I'll make a follow-up PR to core once this is merged, to use the shared action.

[US156216](https://rally1.rallydev.com/#/?detail=/userstory/719304972319&fdp=true)

### DIFF
--- a/comment-on-pr/README.md
+++ b/comment-on-pr/README.md
@@ -1,0 +1,35 @@
+# Comment On PR Action
+
+This GitHub action uses the [GitHub REST API](https://octokit.github.io/rest.js/v19#issues-create-comment) to add a comment to the current PR as the `github-actions` user, with an option to only post once per PR.
+
+## Using the Action
+
+This action **can only be used on `pull_request` event types**, since it posts to the PR that triggers the action. It can be used to write any comment, and can be combined with other steps in your workflow to convey information from your CI process.
+
+In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments.
+
+Here's a sample workflow:
+
+```yml
+name: Comment
+on: pull_request
+jobs:
+  comment:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


comment-on-pr/README.md
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Comment on PR
+        uses: BrightspaceUI/actions/comment-on-pr@main
+        with:
+          MESSAGE: |
+            Thanks for the PR! 🎉
+
+            This is a sample multiline comment.
+          ONLY_POST_ONCE: true
```
svanherk, line 28:
```suggestion
          message: |
            Thanks for the PR! 🎉

            This is a sample multiline comment.
          only-post-once: true
```

comment-on-pr/action.yml
```diff
@@ -0,0 +1,41 @@
...
+    description: A unique key attached invisibly to the comment, for use with the `only-post-once` option
+    default: gh-actions_comment-on-pr
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Find previous comment
+      if: ${{ inputs.only-post-once != 'false' }}
+      uses: Brightspace/third-party-actions@peter-evans/find-comment
```
eKoopmans, line 22: @svanherk suggested moving off of this dependency, but I didn't see an easy way to do that. The closest I saw was [Octokit `issues.listComments`](https://octokit.github.io/rest.js/v19#issues-list-comments) - I could do that and then do a find in [the response](https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#list-issue-comments), searching the `body`.

Would probably work but I didn't want to fiddle with something that already works...

oelhanafey (approved):
LGTM, Gaudi will need to approve adding it

oelhanafey:
Or not, I thought it would need a CODEOWNER approval. Might be good to wait for their approval anyway

dlockhart (approved):
Looks good -- thanks for production-izing this!

PR #121, opened 2023-08-10
US152262 - Add step to open PR and add reviewers

svanherk:
This is heavily based on old `visual-diff`'s [handle-pr.js](https://github.com/BrightspaceUI/actions/blob/82465b48fda225c31eb8850c570567c1a89ea3d6/visual-diff/handle-pr.js). Not tied to the PR name, body, label (if we even want to keep doing that), etc.

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -10,6 +10,9 @@ inputs:
   aws-session-token:
     description: Session token for the role that will assume the vdiff role
     required: true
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Seems good! Haha does anyone use the labels? I guess they could be handy.

bearfriend (approved):

PR #122, opened 2023-08-11
US152262 - Create and Update Commit Status

svanherk:
This replaces the [old comment behavior](https://github.com/BrightspaceUI/actions/blob/main/visual-diff/handle-pr.js#L103-L110) and the [handle-issues.js](https://github.com/BrightspaceUI/actions/blob/26e370ae1a8219e14364a4082c0f9a1a8637db21/visual-diff/handle-issues.js) comments.

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -322,7 +322,7 @@ runs:
         github-token: ${{ inputs.github-token }}
         script: |
           console.log('\x1b[34mOpening Pull Request (if necessary)');
-          if (process.env.TEST_PASSED === 'true' || process.env.GOLDENS_EMPTY === 'true' || process.env.GOLDENS_CONFLICT === 'true') return;
+          if (process.env.TESTS_PASSED === 'true' || process.env.GOLDENS_EMPTY === 'true' || process.env.GOLDENS_CONFLICT === 'true') return;
 
           const prNum = process.env.PULL_REQUEST_NUM;
           const sourceBranchName = process.env.SOURCE_BRANCH;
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -303,7 +322,7 @@ runs:
         github-token: ${{ inputs.github-token }}
         script: |
           console.log('\x1b[34mOpening Pull Request (if necessary)');
-          if (process.env.TEST_PASSED === 'true' || process.env.GOLDENS_EMPTY === 'true' || process.env.GOLDENS_CONFLICT === 'true') return;
+          if (process.env.TESTS_PASSED === 'true' || process.env.GOLDENS_EMPTY === 'true' || process.env.GOLDENS_CONFLICT === 'true') return;
```
svanherk, line 325: Unrelated fix - gives away that I haven't really tested the success case much, just every possible error workflow 😅

vdiff/action.yml
```diff
@@ -58,6 +58,25 @@ runs:
...
+        script: |
+          console.log('\x1b[34mCreating Pending Commit Status');
+          
+          await github.rest.repos.createCommitStatus({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            sha: process.env.ORIGINAL_SHA,
+            state: 'pending',
+            description: 'Run in progress',
+            context: `${context.workflow} / ${context.job} results (${context.eventName})`
```
svanherk, line 74: I can't get the job name, just the job id. But we can set up the examples to set the id only, and if so it'll look like this:
image

dlockhart (approved):
Love it!

bearfriend (approved):

PR #123, opened 2023-08-11
US152262 - Add option to only log errors to publish-to-s3

svanherk:
The S3 sync command is pretty noisy, especially with compression. Publishing the report is only a small piece of `vdiff` and we don't really need to care about each file uploaded.
This PR:
- Adds an option to only log the errors from AWS S3 sync commands in the `publish-to-s3` action
- Use it in the `vdiff` action


dlockhart (approved):
Nice!

PR #126, opened 2023-08-17
Add note about favouring semantic-release

eKoopmans:
I believe we've talked about phasing out this action in favour of `semantic-release`... correct me if I'm wrong! 😄 

This is also mentioned in [How to publish a repo to npm or CodeArtifact](https://desire2learn.atlassian.net/wiki/spaces/DEVCENTRAL/pages/3609003416/How+to+publish+a+repo+to+npm+or+CodeArtifact).

### DIFF
--- a/incremental-release/README.md
+++ b/incremental-release/README.md
@@ -2,6 +2,8 @@
 
 This GitHub action looks for keywords in the latest commit to automatically increment the package version, create a GitHub release/tag, and optionally publish to NPM.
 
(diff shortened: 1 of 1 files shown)


svanherk:
🤔 I think why people like `incremental-release` is the default bump option that `semantic-release` doesn't have. So maybe a better replacement is to recommend `match-lms-release`, _if_ they're tied to LMS version anyways (larger, internal apps)? But feels weird to reference that action here, so maybe this is fine!

Would be interested in other people's thoughts, I do think that ideally everyone is using `semantic-release` for libraries (no default, thoughtful control over the versions) and `match-lms-release` when they don't need to care

PR #129, opened 2023-09-13
Add support for setting files to public-read when syncing to S3

grant-cleary:
Giving a shot at adding support for `--acl public-read` to the `publish-to-s3` action so it can be used to publish files to the CDN. This is a bit out of my wheelhouse so hopefully this isn't crazy broken. 😅

### DIFF
--- a/publish-to-s3/README.md
+++ b/publish-to-s3/README.md
@@ -53,6 +53,7 @@ Options:
 * `cache` (default: `""`): An optional comma-separated list of all file extensions you wish to have cached for 1 year (e.g. `"js,css"`)
 * `cache-default` (default: `""`): An optional default caching policy to apply to all files (e.g. `"--cache-control max-age=120"`)
 * `only-log-errors` (default: false): Only log the errors from aws s3 sync
+* `public-read` (default: false) Sets access permissions on all files to public-read
 
 ## Setting Up AWS Access Creds
 
--- a/publish-to-s3/README.md
+++ b/publish-to-s3/README.md
@@ -53,7 +53,7 @@ Options:
 * `cache` (default: `""`): An optional comma-separated list of all file extensions you wish to have cached for 1 year (e.g. `"js,css"`)
 * `cache-default` (default: `""`): An optional default caching policy to apply to all files (e.g. `"--cache-control max-age=120"`)
 * `only-log-errors` (default: false): Only log the errors from aws s3 sync
-* `public-read` (default: false) Sets access permissions on all files to public-read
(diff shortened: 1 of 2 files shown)


publish-to-s3/action.yml
```diff
@@ -19,6 +19,9 @@ inputs:
   only-log-errors:
     description: Only log the errors from aws s3 sync
     default: false
+  public-read:
```
grant-cleary, line 22: An alternative would be to allow any of the possible ACL options here, but for now I opted for the option most limited in scope. Open to changing this if others disagree.

publish-to-s3/action.yml
```diff
@@ -71,11 +74,12 @@ runs:
         set -f
 
         if [[ ${ONLY_LOG_ERRORS} == "true" ]]; then LOG_LEVEL="--only-show-errors "; fi
+        if [[ ${PUBLIC_READ} == "true" ]]; then ACCESS_PERMISSIONS="--acl public-read ";fi
```
grant-cleary, line 77: Pretty sure the default for `ACCESS_PERMISSIONS` when embedded in a string should be the empty string, but if I'm misunderstanding how this default behaviour works for LOG_LEVEL above, I can also explicitly assign a default.

eKoopmans: Yep should be fine, bash variables default to the empty string.

Nit:

```suggestion
        if [[ ${PUBLIC_READ} == "true" ]]; then ACCESS_PERMISSIONS="--acl public-read "; fi
```

publish-to-s3/README.md
```diff
@@ -53,6 +53,7 @@ Options:
 * `cache` (default: `""`): An optional comma-separated list of all file extensions you wish to have cached for 1 year (e.g. `"js,css"`)
 * `cache-default` (default: `""`): An optional default caching policy to apply to all files (e.g. `"--cache-control max-age=120"`)
 * `only-log-errors` (default: false): Only log the errors from aws s3 sync
+* `public-read` (default: false) Sets access permissions on all files to public-read
```
eKoopmans, line 56: Nit:
```suggestion
* `public-read` (default: false): Sets access permissions on all files to public-read
```

eKoopmans (approved):
The code LGTM - no opinion on what options we want to support!

publish-to-s3/action.yml
dlockhart: This seems good to me.

dlockhart (approved):
Nice!

PR #130, opened 2023-09-21
Add GitHub example for breaking changes

eKoopmans:
Every time I need to do a breaking change I stress about whether I'm doing it right, I'm hoping this helps!


dlockhart (approved):
Haha yes this stresses me out too. Thanks!

PR #131, opened 2023-09-28
Grab job display name instead of id

devpow112:
This uses a round about way to get the actual job display name for the vdiff check status. Right now it uses the `job_id` string value as defined in the [workflow file](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id) not the actual display name. See [below comment](https://github.com/BrightspaceUI/actions/pull/131#discussion_r1340567593) for more details.

Tested with https://github.com/BrightspaceUI/core/pull/4145 and https://github.com/Brightspace/d2l-outcomes/pull/812.

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -58,24 +58,52 @@ runs:
         PREFIX: ${{ inputs.vdiff-branch-prefix }}
       shell: bash
 
+    - name: Get Job Name
+      id: job-name
+      uses: Brightspace/third-party-actions@actions/github-script
+      with:
+        github-token: ${{ inputs.github-token }}
+        result-encoding: string
+        script: |
+          console.log('\x1b[34mGetting Job Name');
+
+          const { data: { jobs } = {} } = await github.rest.actions.listJobsForWorkflowRunAttempt({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            run_id: context.runId,
+            attempt_number: process.env.GITHUB_RUN_ATTEMPT
+          });
+          const runnerName = process.env.RUNNER_NAME;
+          const jobNames = jobs
+            .filter(({runner_name}) => runner_name === runnerName)
+            .map(({name}) => name);
+
+          return jobNames.length === 1 ? jobNames[0] : context.job;
+      env:
+        FORCE_COLOR: 3
+
     - name: Create Commit Status
       uses: Brightspace/third-party-actions@actions/github-script
       with:
         github-token: ${{ inputs.github-token }}
         script: |
           console.log('\x1b[34mCreating Pending Commit Status');
-          
+
+          const jobName = process.env.JOB_NAME;
+          const resultsText = jobName[0].toUpperCase() === jobName[0] ? 'Results' : 'results'
+
           await github.rest.repos.createCommitStatus({
             owner: context.repo.owner,
             repo: context.repo.repo,
             sha: process.env.ORIGINAL_SHA,
             state: 'pending',
             description: 'Run in progress',
-            context: `${context.workflow} / ${context.job} results (${context.eventName})`
+            context: `${context.workflow} / ${jobName} ${resultsText} (${context.eventName})`
           });
       env:
         FORCE_COLOR: 3
         ORIGINAL_SHA: ${{ steps.run-info.outputs.original-sha }}
+        JOB_NAME: ${{ steps.job-name.outputs.result }}
 
     - name: vdiff Branch Cleanup
       uses: Brightspace/third-party-actions@actions/github-script
@@ -103,7 +131,7 @@ runs:
                 repo: context.repo.repo,
                 pull_number: prNum
               });
-        
+
               if (prInfo.data.state !== 'open') {
                 prOpen = false;
               }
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -58,24 +58,52 @@ runs:
...
+          });
+          const runnerName = process.env.RUNNER_NAME;
+          const jobNames = jobs
+            .filter(({runner_name}) => runner_name === runnerName)
+            .map(({name}) => name);
+
+          return jobNames.length === 1 ? jobNames[0] : context.job;
+      env:
+        FORCE_COLOR: 3
+
```
devpow112, line 84: This isn't always guaranteed to work since `RUNNER_NAME` isn't guaranteed to be unique, but there is a high chance the intersection of run attempt, run, repo and workflow are pretty unique so it should work in most cases. In the event we get more then one result it falls back to using the `context.job` which will be the string job id defined in the workflow file.

vdiff/action.yml
```diff
@@ -58,24 +58,52 @@ runs:
...
     - name: Create Commit Status
       uses: Brightspace/third-party-actions@actions/github-script
       with:
         github-token: ${{ inputs.github-token }}
         script: |
           console.log('\x1b[34mCreating Pending Commit Status');
-          
+
+          const jobName = process.env.JOB_NAME;
+          const resultsText = jobName[0].toUpperCase() === jobName[0] ? 'Results' : 'results'
```
devpow112, line 93: This is just my brain bugging me. For some runs we have stuff like `Visual Diff` so it's nicer to me to show `Visual Diff Results` to match the casing and in other cases we have `vdiff` so `vdiff results` make my brain happy. I'm find to remove this if it's dumb just added it to see what it looked like and make my brain shut up 😛 .

vdiff/action.yml
```diff
@@ -483,14 +511,17 @@ runs:
             core.setFailed('Completed - Build Failed.');
           }
 
+          const jobName = process.env.JOB_NAME;
+          const resultsText = jobName[0].toUpperCase() === jobName[0] ? 'Results' : 'results'
```
devpow112, line 515: This is the same comment as above. Can remove if we think it's overkill.

svanherk:
Lol I was really hoping to get away with id because I was so annoyed at how crazy complicated GitHub makes getting the name.  Is the reason you can't use the id (and not set a name at all) because you want spaces?

devpow112:
> Lol I was really hoping to get away with id because I was so annoyed at how crazy complicated GitHub makes getting the name. Is the reason you can't use the id (and not set a name at all) because you want spaces?

I can it just wouldn't match any of our current set-up. I'm not opposed to not doing this it was just bugging me so wanted to see if was possible. If it's too much I can abandon it @svanherk.

svanherk:
> > Lol I was really hoping to get away with id because I was so annoyed at how crazy complicated GitHub makes getting the name. Is the reason you can't use the id (and not set a name at all) because you want spaces?
> 
> I can it just wouldn't match any of our current set-up. I'm not opposed to not doing this it was just bugging me so wanted to see if was possible. If it's too much I can abandon it @svanherk.

Mostly interested in the reasoning because I also investigated and then backtracked and called it because I thought I was being picky 😝. If it seems solid and quick I don't mind keeping it

vdiff/action.yml
svanherk: Might as well move it into the actual calculation of `jobName`, since we don't actually use it without the "Results" part?

svanherk: Could change it to "Get Status Check Name" to be more accurate

devpow112:
> > > Lol I was really hoping to get away with id because I was so annoyed at how crazy complicated GitHub makes getting the name. Is the reason you can't use the id (and not set a name at all) because you want spaces?
> > 
> > 
> > I can it just wouldn't match any of our current set-up. I'm not opposed to not doing this it was just bugging me so wanted to see if was possible. If it's too much I can abandon it @svanherk.
> 
> Mostly interested in the reasoning because I also investigated and then backtracked and called it because I thought I was being picky 😝. If it seems solid and quick I don't mind keeping it

The main motivation for me is a preference thing. I want to be able to call my jobs whatever I want and have the job name's line up. Personally I like them being more title case but again that's just a preference. I doubt many other people care 😝. What I implemented is based on all the suggestions I found. It should work 100% for self hosted since each runner name is unique the not 100% case would be for public runners but it's a small chance it won't work for those. In worst case it just falls back into using the `context.job` which is what it was doing before anyways so it's no worse in the worst case. I find it weird how inconsistent some of the GitHub job ID related stuff is. They use the ID in the workflow and the numeric DB ID interchangeably in a lot of discussions/documentation which seems to cause a lot of confusion.

svanherk:
> I find it weird how inconsistent some of the GitHub job ID related stuff is. They use the ID in the workflow and the numeric DB ID interchangeably in a lot of discussions/documentation which seems to cause a lot of confusion.

In general, GitHub is weird with IDs.  I forget the exact case, but to do something (maybe turn on automerge?) I had to use GraphQL to get a special database ID that you could _only_ get there.

devpow112:
I think unless you have any objects @svanherk I'll go ahead with this.

svanherk:
> I think unless you have any objects @svanherk I'll go ahead with this.

Yup I'm good with it, thoughts on https://github.com/BrightspaceUI/actions/pull/131#discussion_r1340582020?

vdiff/action.yml
devpow112: Ah yes, sorry missed this. I'll adjust that. Makes sense to not do it twice. Maybe I'll just change that to output the check name like you said

devpow112:
Retested with https://github.com/Brightspace/d2l-outcomes/pull/814.

svanherk (approved):
Nice, thanks!

PR #132, opened 2023-10-10
US157919 - Handle dependabot

svanherk:
If the actor is `dependabot`, we now skip the "nice-to-have" results commit status.  This allows passing tests to go green with no issue. If the tests fail, the opening of the goldens PR will fail unless additional permissions have been configured in the workflow.  A warning has been added about that.

This PR also fixes the username and email for the bot, which is unrelated.


dlockhart (approved):
🎉

bearfriend (approved):

PR #138, opened 2024-01-19
Adding publish-to-cdn action (GAUD-726)

dlockhart:
This introduces a new action specifically for publishing to the Brighspace CDN. We noticed that in almost all the new uses of `publish-to-s3`, caching of assets was desired by wasn't configured. This hopefully resolves that by tuning it specially for the CDN.

Additionally, this adds overwrite protection which [was present in the old frau-publisher](https://github.com/Brightspace/frau-publisher/blob/master/src/overwrite.js) to prevent overwriting existing things on the CDN.

### DIFF
--- a/publish-to-cdn/README.md
+++ b/publish-to-cdn/README.md
@@ -1,0 +1,66 @@
+# Publish To CDN Action
+
+This action leverages the [publish-to-s3](../publish-to-s3/) GitHub Action, and configures it specifically for publishing to the [Brightspace CDN](https://desire2learn.atlassian.net/wiki/spaces/DEVCENTRAL/pages/3314221651/Brightspace+CDN) (`https://s.brightspace.com`).
+
+Features:
+- Prevents existing files from being overwritten
+- Caches all assets for 1 year
+- Enables `public-read` access
+- Prefixes `bucket-path` with `s3://d2lprodcdn/`
+
+> [!IMPORTANT]  
+> Published files will be cached indefinitely, so ensure a mechanism for cache invalidation is present, such as a version number in the `cdn-path`. For example: `my-lib/<version>`.
+
+## Using the Action
+
+This action is typically triggered from a release workflow that runs on your `main` branch when a new release is created.
+
+To assume the correct role for publishing to the CDN and have the appropriate secrets available, set up the `cdn` capability [in repo-settings](https://github.com/Brightspace/repo-settings/blob/main/docs/cdn.md).
+
+Here's a sample workflow:
+
+```yml
+name: Publish
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish:
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+      - name: Setup Node
+        uses: Brightspace/third-party-actions@actions/setup-node
+        with:
+          node-version-file: .nvmrc
+      - name: Install dependencies
+        run: npm install
+      - name: Release
+        id: release
+        uses: <semantic/match-lms/incremental>
+      - name: Assume role
+        if: steps.release.outputs.VERSION != ''
+        uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
+          role-to-assume: <role_that_can_upload_to_the_cdn>
+          role-duration-seconds: 3600
+          aws-region: ca-central-1
+      - name: Publish
+        if: steps.release.outputs.VERSION != ''
+        uses: BrightspaceUI/actions/publish-to-cdn@main
+        with:
+          cdn-path: my-lib/${{ steps.release.outputs.VERSION }}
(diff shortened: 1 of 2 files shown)


publish-to-cdn/README.md
```diff
@@ -0,0 +1,66 @@
+# Publish To CDN Action
+
+This action leverages the [publish-to-s3](../publish-to-s3/) GitHub Action, and configures it specifically for publishing to the [Brightspace CDN](https://desire2learn.atlassian.net/wiki/spaces/DEVCENTRAL/pages/3314221651/Brightspace+CDN) (`https://s.brightspace.com`).
```
dlockhart, line 3: I know this is a public repo and this doc is internal, but I feel like there aren't enough links between all the different places that talk about this stuff so felt it was worth it. _Probably_ this repo should be private.

publish-to-cdn/action.yml
```diff
@@ -0,0 +1,31 @@
...
+           exit 1;
+        fi
+      shell: bash
+      env:
+        CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Deploy
+      uses: BrightspaceUI/actions/publish-to-s3@main
+      with:
+        bucket-path: s3://d2lprodcdn/${{ inputs.cdn-path }}
+        cache-default: --cache-control public,max-age=31536000,immutable
```
dlockhart, line 28: Matches this:
https://github.com/Brightspace/frau-publisher/blob/master/src/s3.js#L61

publish-to-cdn/action.yml
```diff
@@ -0,0 +1,31 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Overwrite Protection
+      run: |
+        if [[ $(aws s3 ls "s3://d2lprodcdn/${CDN_PATH}/" | head) ]]; then
+           echo "ERROR: cdn-path "\""$CDN_PATH"\"" already exists, which would result in files being overwritten. Choose a unique publish path.";
+           exit 1;
+        fi
```
dlockhart, line 20: I've tested this out in the [fonts-lib repo here](https://github.com/Brightspace/fonts-lib/pull/63).

eKoopmans (approved):
LGTM!

publish-to-cdn/README.md
svanherk: Is this possible? 🤔 Not sure I've tried to access actions from a private repo in a public one (like `core`)

dlockhart: Oh I'm sure it isn't -- it _should_ be private since there are private-y things like the CDN in here, but realistically can't.

svanherk: A lot of our other action READMEs point to internal stuff to discuss CodeArtifact, so this definitely isn't the first spot that has to do it

publish-to-cdn/README.md
```diff
@@ -0,0 +1,66 @@
...
+- Caches all assets for 1 year
+- Enables `public-read` access
+- Prefixes `bucket-path` with `s3://d2lprodcdn/`
+
+> [!IMPORTANT]  
+> Published files will be cached indefinitely, so ensure a mechanism for cache invalidation is present, such as a version number in the `cdn-path`. For example: `my-lib/<version>`.
+
+## Using the Action
+
+This action is typically triggered from a release workflow that runs on your `main` branch when a new release is created.
```
svanherk, line 16: Are we still trying to avoid "your" terminology (versus saying "the `main` branch")?

publish-to-cdn/README.md
```diff
@@ -0,0 +1,66 @@
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
+          role-to-assume: <role_that_can_upload_to_the_cdn>
+          role-duration-seconds: 3600
+          aws-region: ca-central-1
+      - name: Publish
+        if: steps.release.outputs.VERSION != ''
+        uses: BrightspaceUI/actions/publish-to-cdn@main
+        with:
+          cdn-path: my-lib/${{ steps.release.outputs.VERSION }}
```
svanherk, line 59:
```suggestion
          cdn-path: <lib-name>/${{ steps.release.outputs.VERSION }}
```
Or something like that, mostly adding the `<` and `>` to imply they need to replace it like the other spots above

publish-to-cdn/action.yml
```diff
@@ -0,0 +1,31 @@
...
+      shell: bash
+      env:
+        CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Deploy
+      uses: BrightspaceUI/actions/publish-to-s3@main
+      with:
+        bucket-path: s3://d2lprodcdn/${{ inputs.cdn-path }}
+        cache-default: --cache-control public,max-age=31536000,immutable
+        publish-directory: ${{ inputs.publish-directory }}
+        only-log-errors: true
```
svanherk, line 30: I'm torn on this one. I added it to avoid crazy long logs for uploading vdiff reports, preview sites, etc. but maybe we like seeing the logs of everything uploaded to prod?

publish-to-cdn/README.md
dlockhart: Haha I am -- thanks.

publish-to-cdn/action.yml
dlockhart: Hmm yeah agree good call -- updated.

svanherk: If we do find we want it off for certain scenarios, we can always add an input to configure it later

svanherk (approved):

PR #139, opened 2024-01-19
automatically assume the role (GAUD-726)

dlockhart:
Based on @j3parker's [suggestion here](https://github.com/Brightspace/repo-settings/pull/1575).

If the `771734770799` IAM is different based on the org (BrightspaceUI, Brightspace, BrightspaceHypermediaComponents, etc.) then this won't work.

### DIFF
--- a/publish-to-cdn/action.yml
+++ b/publish-to-cdn/action.yml
@@ -30,13 +30,24 @@ runs:
       shell: bash
       env:
         CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Determine Role Name
+      id: role-name
+      run: |
+        ORG_NAME="${{ github.repository_owner }}"
+        if [[ ${ORG_NAME} == "BrightspaceHypermediaComponents" ]]; then ORG_NAME="BHC"; fi
+
+        REPO_NAME="${{ github.repository }}"        
+        REPO_NAME="${REPO_NAME#*/}"
+
+        echo "value=arn:aws:iam::771734770799:role/r+$ORG_NAME+$REPO_NAME+repo" >> $GITHUB_OUTPUT
+      shell: bash
     - name: Assume role
       uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
       with:
         aws-access-key-id: ${{ inputs.aws-access-key-id }}
         aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
         aws-session-token: ${{ inputs.aws-session-token }}
-        role-to-assume: "arn:aws:iam::771734770799:role/r+${{ github.repository }}+repo"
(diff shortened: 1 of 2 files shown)


publish-to-cdn/action.yml
```diff
@@ -21,6 +21,15 @@ runs:
       shell: bash
       env:
         CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Assume role
+        uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
```
svanherk, line 29: I think these need to be inputs they pass in, the action can't access them directly?

publish-to-cdn/action.yml
```diff
@@ -21,6 +30,15 @@ runs:
       shell: bash
       env:
         CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Assume role
+      uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+      with:
+        aws-access-key-id: ${{ inputs.aws-access-key-id }}
+        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
+        aws-session-token: ${{ inputs.aws-session-token }}
+        role-to-assume: "arn:aws:iam::771734770799:role/r+${{ github.repository }}+repo"
```
svanherk, line 39: We'll need to swap the `/` for a `+` in `github.repository`, and [maybe do something special for the HM org](https://github.com/Brightspace/repo-settings/pull/1575#issuecomment-1901168054).

dlockhart: Thanks.  I've added an extra step to compute this... local testing _seems_ to work but who knows until it actually runs.

publish-to-cdn/action.yml
```diff
@@ -21,6 +30,26 @@ runs:
...
       env:
         CDN_PATH: ${{ inputs.cdn-path }}
+    - name: Determine Role Name
+      id: role-name
+      run: |
+        ORG_NAME="${{ github.repository_owner }}"
+        if [[ ${ORG_NAME} == "BrightspaceHypermediaComponents" ]]; then ORG_NAME="BHC"; fi
+
+        REPO_NAME="${{ github.repository }}"        
+        REPO_NAME="${REPO_NAME#*/}"
```
dlockhart, line 40: This supposedly strips everything before the `/`.

svanherk (approved):

PR #142, opened 2024-02-22
set PUPPETEER_DOWNLOAD_BASE_URL to fix Chrome

dlockhart:
Chrome [moved where its downloads are](https://github.com/GoogleChromeLabs/chrome-for-testing/commit/dd90f461197402d01208a690b58f8a7776db6e7c), which is causing the older version of Puppeteer that we're still using to fail to find it.

This overrides the location to the new one, which will hopefully fix things until we EOL this library.

### DIFF
--- a/visual-diff/action.yml
+++ b/visual-diff/action.yml
@@ -33,6 +33,7 @@ runs:
         echo -e "\e[34mInstalling Dependencies"
         NPM_PACKAGE=$(cut -d "=" -f 2 <<< $(npm run env | grep "npm_package_name"))
         if [ $NPM_PACKAGE != '@brightspace-ui/visual-diff' ]; then
+          PUPPETEER_DOWNLOAD_BASE_URL=https://storage.googleapis.com/chrome-for-testing-public
           npm install @brightspace-ui/visual-diff@14 --no-save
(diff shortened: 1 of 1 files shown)


visual-diff/action.yml
```diff
@@ -33,6 +33,7 @@ runs:
         echo -e "\e[34mInstalling Dependencies"
         NPM_PACKAGE=$(cut -d "=" -f 2 <<< $(npm run env | grep "npm_package_name"))
         if [ $NPM_PACKAGE != '@brightspace-ui/visual-diff' ]; then
+          PUPPETEER_DOWNLOAD_BASE_URL=https://storage.googleapis.com/chrome-for-testing-public
```
dlockhart, line 36: I _think_ this is the right place... this won't run in the actual `@brightspace-ui/visual-diff` repo so if we care about it working we'd need to add this line there as well before it installs its dependencies.

bearfriend (approved):

svanherk (approved):
Let's see how it goes - we can rerun https://github.com/BrightspaceUI/documentation/actions/runs/8009457834/job/21878153671?pr=1578 to test it works

PR #146, opened 2024-02-23
Fix reading directory as file warning

devpow112:
Noticed this while watching a copy to s3
```
failed to read input [./_site/internal/tags/node.js]: Is a directory
```
Since the folder ends in `.js` it tries to add it to the list of files to compress. This should make `find` only look for `file` types.


dlockhart (approved):
Good catch, thanks!

PR #147, opened 2024-03-05
Add ability to have updating comment

devpow112:
Would be nice to make this part of the action. Right now we have a number of workflows that hid manually then repost using this action.

### DIFF
--- a/comment-on-pr/README.md
+++ b/comment-on-pr/README.md
@@ -6,9 +6,12 @@ This GitHub action uses the [GitHub REST API](https://octokit.github.io/rest.js/
 
 This action **can only be used on `pull_request` event types**, since it posts to the PR that triggers the action. It can be used to write any comment, and can be combined with other steps in your workflow to convey information from your CI process.
 
-In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments.
+In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments. If you want to keep posting updated content using `hid-previous` will make sure to hide your previous comment before posting new content.
 
-Here's a sample workflow:
+> [!NOTE]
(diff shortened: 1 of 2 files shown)


comment-on-pr/README.md
```diff
@@ -6,9 +6,12 @@ This GitHub action uses the [GitHub REST API](https://octokit.github.io/rest.js/
 
 This action **can only be used on `pull_request` event types**, since it posts to the PR that triggers the action. It can be used to write any comment, and can be combined with other steps in your workflow to convey information from your CI process.
 
-In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments.
+In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments. If you want to keep posting updated content using `hid-previous` will make sure to hide your previous comment before posting new content.
```
dlockhart, line 9:
```suggestion
In situations where you only want to post once on a given PR, use `only-post-once: true`. Every comment from this action includes a hidden key in an HTML comment, which is used by `only-post-once` to detect if a post has already been created. Use `unique-key` to specify your own unique key, which can be useful e.g. if you want to uniquely post two different comments.

To keep posting updated content, `hide-previous: true` hides previous comments before posting new content.
```

comment-on-pr/README.md
```diff
@@ -6,9 +6,12 @@ This GitHub action uses the [GitHub REST API](https://octokit.github.io/rest.js/
...
 
-Here's a sample workflow:
+> [!NOTE]
+> The `only-post-once` and `hide-previous` can not be used together. If you use `only-post-once` and `hide-previous` the comment will only be posted once.
```
dlockhart, line 12: Too bad `only-post-once` is used in so many places... otherwise it would have been nice to have this be an enum.

devpow112: Yeah I was thinking about that. I could ad the new property as a mode and just update it everywhere. I'm doing large scale action updates for some other stuff so I could take care of this at the same time as some of the others.

devpow112: Maybe `post-mode: once / update / always`?

dlockhart: I like the `post-mode` option for sure... but yeah only if you feel like switching everything!

devpow112: I'm fine with updating it. Let me add a new post mode that will work with the current `only-post-once`. Then we can update stuff. Not gonna get to that today. Will try in the next few days. Will close this for now.

PR #150, opened 2024-03-06
Add pull request number

devpow112:
I'm mainly adding this because of workflows that trigger off issue/pull request comments but it could be useful to have the ability to run this and post comments to another PR. Without this I'd have to do it all myself in each workflow and I don't want to 😛. Tested on  https://github.com/Brightspace/playwright-ui-automation/pull/1067, [this run](https://github.com/Brightspace/playwright-ui-automation/actions/runs/8179235976/job/22364874598) had it manually specified, and [this run](https://github.com/Brightspace/playwright-ui-automation/actions/runs/8179253212/job/22364930752) didn't include it at all but was within a pull request context so the default resolves to something.

### DIFF
--- a/comment-on-pr/action.yml
+++ b/comment-on-pr/action.yml
@@ -7,16 +7,16 @@ inputs:
   only-post-once:
     description: Check whether the current message has already been posted before posting again
     default: false
-    required: true
+    required: false
     deprecationMessage: 'Please use the input `post-mode: once` instead, this input will be going away in the future'
   post-mode:
     description: 'How posting of the comment will be handled. Options are `always`, `once`, `hide-previous` or `update`'
     default: always
-    required: true
+    required: false
   unique-key:
     description: 'A unique key attached invisibly to the comment, for use with the `only-post-once` option'
     default: gh-actions_comment-on-pr
-    required: true
+    required: false
   pull-request-number:
     description: The number of the pull request to post the comment to
     default: ${{ github.event.pull_request.number }}


comment-on-pr/action.yml
```diff
@@ -7,16 +7,20 @@ inputs:
...
-    required: false
+    required: true
   unique-key:
     description: 'A unique key attached invisibly to the comment, for use with the `only-post-once` option'
     default: gh-actions_comment-on-pr
-    required: false
+    required: true
+  pull-request-number:
+    description: The number of the pull request to post the comment to
+    default: ${{ github.event.pull_request.number }}
```
devpow112, line 22: In a run where you have no pull request context this just ends up empty which triggers the required error flow.

comment-on-pr/action.yml
```diff
@@ -7,16 +7,20 @@ inputs:
   only-post-once:
     description: Check whether the current message has already been posted before posting again
     default: false
-    required: false
+    required: true
```
dlockhart, line 10: Did you mean to flip this?

devpow112: No. I didn't mean to flip the other one for `post-mode` either. I'll revert those.

devpow112: Actually this stops you from doing something like this I believe
```yml
only-post-once: ''
```
Might be good to keep

devpow112: I'll revert it just in case. Might test that out though

dlockhart (approved):
Cool!

devpow112:
I'm gonna wait to merge this tomorrow. Don't want to merge now and have stuff broken overnight. Don't think it will be a problem but you never know 🤷‍♂️

PR #151, opened 2024-03-06
Remove `only-post-once` from documentation

devpow112:
[Last use of it was removed](https://github.com/search?q=%22BrightspaceUI%2Factions%2Fcomment-on-pr%22+%22only-post-once%22+NOT+repo%3ABrightspaceUI%2Factions+NOT+repo%3ABrightspace%2Fblog.d2l.dev+NOT+repo%3ABrightspaceGraveyard%2Ftest-reporting.d2l.dev&type=code). Leaving option in for now. Don't want to break recent release branches that may use the old option. Seems good to not advertise it though. Will remove fully after a few release.


dlockhart (approved):
Sounds like a plan, thanks!

PR #154, opened 2024-04-01
Add easy-to-copy link to the CDN file location

svanherk:
`frau-publisher` used to do something like this. Just gives an easy way to copy the public path to the CDN.


dlockhart (approved):
Nice!

PR #156, opened 2024-04-03
GAUD-6207 - Add slack notifications to update-package-lock

svanherk:
Add inputs to `update-package-lock` to tell the action to:
- Optionally send a slack message to the channel of your choice when it fails
- Optionally send a slack message to the channel of your choice when it pushes to an `update-package-lock` PR that has been open for more than three days
  - This is useful if you have turned off CODEOWNER review for these PRs to auto-merge themselves
  - CI in the repo might be flaky and this PR might be about to pass CI and merge itself, but we don't have a great way to catch actual dependency changes that are breaking something, or a PR that GitHub silently failed to turn on auto-merge for

I've tested this in `repo-data` and `fra-dashboard` pointing to this branch. Here's what the messages look like:
image


dlockhart (approved):
So cool!

PR #157, opened 2024-04-03
GAUD-6207 - Add note about needing to set up the slack app in the channel

svanherk:
In case someone hypothetically forgets to add the app to the channel... 😅


dlockhart (approved):
😆

PR #158, opened 2024-04-04
Add comment id as output to comment on PR action

devpow112:
This is useful mainly for testing. I have action that internally use this action and I want to validate that they generate the correct comment. Having the ID means I can easily grab the comment my action made and validate it's contents. Validated with https://github.com/Brightspace/test-actions/pull/44.


dlockhart (approved):
🆒

PR #162, opened 2024-06-03
Improve display of dependency changes when exceeding max length

eKoopmans:
We [had a PR](https://github.com/Brightspace/parent-portal-ui/pull/1417) with no summary of changes, and I was wondering why we it didn't have one so I checked the code.

I've updated the logic to include a warning and as much of the changes as possible, when the change-list exceeds the max length.

### DIFF
--- a/update-package-lock/handle-pr.js
+++ b/update-package-lock/handle-pr.js
@@ -135,7 +135,12 @@ async function handlePR() {
 	let prBody = 'Automatic update of the \`package-lock.json\` file.';
 	const prDependencyDiff = getDependencyDiff();
 
-	if (prDependencyDiff !== '' && (prBody.length + prDependencyDiff.length) < prDescriptionMaxSize) {
+	if ((prBody.length + prDependencyDiff.length) > prDescriptionMaxSize) {
+		prBody += '\n> [!WARNING]\n> Dependency Changes exceed max description length and have been truncated.\n\n';
+		const maxLength = prDescriptionMaxSize - prBody.length;
+		const truncatedLength = prDependencyDiff.lastIndexOf('\n', maxLength);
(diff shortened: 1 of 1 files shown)


update-package-lock/handle-pr.js
```diff
@@ -135,7 +135,12 @@ async function handlePR() {
 	let prBody = 'Automatic update of the \`package-lock.json\` file.';
 	const prDependencyDiff = getDependencyDiff();
 
-	if (prDependencyDiff !== '' && (prBody.length + prDependencyDiff.length) < prDescriptionMaxSize) {
+	if ((prBody.length + prDependencyDiff.length) > prDescriptionMaxSize) {
```
eKoopmans, line 138: Note a slight change here, previous code allowed a max length of 65535, new code allows 65536 (which I think is fine?).

update-package-lock/handle-pr.js
```diff
@@ -135,7 +135,12 @@ async function handlePR() {
...
+		prBody += '\n> [!WARNING]\n> Dependency Changes exceed max description length and have been truncated.\n\n';
+		const maxLength = prDescriptionMaxSize - prBody.length;
+		const truncatedLength = prDependencyDiff.lastIndexOf('\n', maxLength);
+		prBody += prDependencyDiff.slice(0, truncatedLength !== -1 ? truncatedLength : maxLength);
```
eKoopmans, line 142: I'm fine with just always truncating at `maxLength` if we prefer, knowing that the last line of the table might look funny.

dlockhart (approved):
Thanks!

eKoopmans:
Oh also this PR will truncate the trailing `</details>`, but I double-checked and things still work with that closing tag missing.

PR #164, opened 2024-07-29
remove old visual diff action

dlockhart:
Nothing is calling this, so removing the action.

The cert period for the August release also ended last Friday, so I think it will be very unlikely that a repo still using the old visual diff would need to hotfix. If they do, they can either:
- Skip the vdiff tests since it's not a critical step
- Backport the vdiff migration
- We could create a branch of this repo before this removal happened and they could point there


devpow112 (approved):
🎉

PR #166, opened 2024-09-16
`vdiff`: Detect orphaned goldens and remove them

bearfriend:
Currently we only create a vdiff PR if one or more tests fail. This means if everything passes but some tests were removed, orphaned goldens will remain in the repo until some subsequent run that has failed vdiff tests.

This detects the orphaned goldens and forces the workflow to procede as if some tests had failed.

[Example result](https://github.com/BrightspaceUI/testing/pull/496/files) - Everything in the report passes but the PR still removes the orphaned goldens.

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -280,8 +280,7 @@ runs:
           mv "${PASS_DIR}/"* "./${TEST_PATH}/golden/${TEST}"
         done
 
-        if [[ ${COUNT_MISMATCH} = true ]]; then
-        echo -e "\e[31mGolden count has changed."
+        if [[ ${COUNT_MISMATCH} = true ]]; then echo -e "\e[31mGolden count has changed."; fi
 
         echo "Moving failed screenshots to golden directories"
         find ./.vdiff -name fail -type d | while read FAIL_DIR; do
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -256,19 +256,31 @@ runs:
...
+            echo "passed=false" >> ${GITHUB_OUTPUT}
+            COUNT_MISMATCH=true
+          fi
+
           mkdir -p "./${TEST_PATH}/golden/${TEST}"
           mv "${PASS_DIR}/"* "./${TEST_PATH}/golden/${TEST}"
-        done
+        done <<< "$(find ./.vdiff -name pass -type d)"
+
+        if [[ $COUNT_MISMATCH = true ]]; then echo -e "\e[31mGolden count has changed"; fi
```
dlockhart, line 283: If you removed a golden and also added a new one, this would still "work" since the newly added golden would "fail" the test run?

dlockhart (approved):
Nice!

vdiff/action.yml
bearfriend: Yep, exactly!

PR #172, opened 2024-12-17
Add additional inputs/flags

bearfriend:
Adds additional standard `mfv` inputs/flags for controlling the CLI, as well as a new `format --correct` flag that was split out from "Essentials" to now be optional ([example](https://github.com/Brightspace/manager-view-fra/pull/2889/commits/a07a0d0852c631a6e695a468aeb95f300f9a5ee5))

### DIFF
--- a/t9n-format/action.yml
+++ b/t9n-format/action.yml
@@ -189,6 +189,7 @@ runs:
     - name: Format Translations - Argument Correction
       if: ${{ inputs.t9n-correct != 'false' }}
       run: |
+        set -x
         echo -e "\e[34mFormatting translations - Argument Correction"
         if [[ "${LOCALES}" != "''" ]]; then FLAGS="--locales ${LOCALES} "; fi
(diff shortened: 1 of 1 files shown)


t9n-format/action.yml
```diff
@@ -20,8 +20,17 @@ inputs:
     description: Replace quote characters with locale-appropriate characters ("source", "straight", or "both")
     default: 'straight'
   t9n-locales:
-    description: Comma-separated list of locales to format
+    description: Comma-separated list of locales to format. If empty, all locales are formatted.
     default: ''
+  t9n-source:
+    description: The locale to use as the source
+    default: ''
```
bearfriend, line 27: This currently defaults to `''` instead of `'en'` so that a config file is not overridden. Sort of flip-flopping if it should just override it anyway. Setting the input once wouldn't be a big deal.

dlockhart (approved):

PR #179, opened 2025-04-17
GAUD-7792 - Remove release-branch-management note

svanherk:
This is no longer required and creating release branches should "just work"!


dlockhart (approved):
🎉

PR #189, opened 2025-06-09
`t9n-format`: Add file formatting steps and docs

bearfriend:
[GAUD-8160](https://desire2learn.atlassian.net/browse/GAUD-8160): `t9n-format` action: Add file formatting steps

Tested in BrightspaceUI/core#5848

[GAUD-8160]: https://desire2learn.atlassian.net/browse/GAUD-8160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/t9n-format/README.md
+++ b/t9n-format/README.md
@@ -65,7 +65,7 @@ A glob path to the directory containing translation files. e.g. `lang/`
 * `t9n-locales`<br>
 A comma-separated list of locales to limit all operations to. By default all files in the `t9n-path` are acted on.
 
-#### Message Formatting:
+#### Message Formatting
 
 * `t9n-newlines` (default: `false`)<br>
 When formatting complex arguments, use newlines and indentation for readability


t9n-format/README.md
```diff
@@ -41,30 +41,63 @@ jobs:
...
+Replace quote characters with locale-appropriate characters ("source", "straight", or "both")
+
+> [!WARNING]
+> Use these options with extreme caution. If you don't understand the implications, you shouldn't use them.
+>
+> * `t9n-add` (default: `false`)<br>
+> Add cases for supported but missing plural and selectordinal categories
+>
+> * `t9n-dedupe` (default: `false`)<br>
+> Remove complex argument cases that duplicate the `other` case. Takes precedence over `t9n-add`.
```
svanherk, line 89: It's pretty common to copy/paste workflows from one repo to another. If `core` uses these, this warning is likely not going to be seen before being used.

bearfriend: Nothing will use these except maybe TIE

bearfriend: They're now both disabled. The reason I want to include the warning here is so that people can get some idea that using the tool locally on on their own in CI, these options are dangerous. But I don't want to disable them in mfv itself.

svanherk: Right, and we need them in the action because TIE will use this action?

bearfriend: Maybe. It may end up that TIE just needs to be updated to handle repos using these directly. It's something I'll have to figure out down the road after also talking more with the translation vedors.

PR #190, opened 2025-06-09
`t9n-format`: Add validate step

bearfriend:
Adds the main `mfv` validation functionality to the action, _before_ any formatting changes happen. Running it before allows for comparison of the issue report from the original PR state with the report generated on the next run after any changes are merged.

Most useful for issues that couldn't be fixed like anything that would prevent message parsing. e.g. `An unclosed {argument`

Test run [here](https://github.com/BrightspaceUI/core/actions/runs/15543900504/job/43760933896?pr=5850#step:5:175)


t9n-format/action.yml
```diff
@@ -188,6 +191,21 @@ runs:
         FORCE_COLOR: 3
         PREFIX: ${{ inputs.t9n-branch-prefix }}
 
+    - name: Validate
+      if: ${{ inputs.t9n-validate != 'false' }}
+      run: |
+        echo -e "\e[34mRunning validation"
+        if [[ "${LOCALES}" != "''" ]]; then FLAGS="--locales ${LOCALES} "; fi
+        if [[ ${SOURCE} != "" ]]; then FLAGS="${FLAGS} --source-locale ${SOURCE} "; fi
+        if [[ ${LOCALES_PATH} != "" ]]; then FLAGS="${FLAGS} --path ${LOCALES_PATH} "; fi
```
svanherk, line 200: I feel like at some point, might be nice to factor this into a step where we calculate it once, output it, and use it in each step lol

bearfriend: Hah, fair point. When it was all being set up there was a lot of iterating and they weren't all the same, but they all have these three lines now, if in a different order.

PR #192, opened 2025-06-26
GAUD-7842 - Output new PR num for vdiff when opened

svanherk:
I'm working on the goldens update process for the `testing` repo, and it will make things cleaner if I'm given this PR number (rather than having to query for it after the fact).


bearfriend (approved):

svanherk:
Closing this, we're going a different direction.

PR #213, opened 2026-04-20
Add support for alt tests

GZolla:
[GAUD-9785](https://desire2learn.atlassian.net/browse/GAUD-9785)

Tested branch in [here](https://github.com/BrightspaceUI/core/pull/6835)

[GAUD-9785]: https://desire2learn.atlassian.net/browse/GAUD-9785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -300,7 +300,14 @@ runs:
             mkdir -p "./${TEST_PATH}/golden/${TEST}/${BROWSER}"
             for PNG in `ls "${FAIL_DIR}/${BROWSER}"`; do
               NAME=`basename "${PNG}"`
-              if  [[ ${NAME} != *"-diff.png" ]] && [[ ${NAME} != *"-resized-golden.png" ]] && [[ ${NAME} != *"-resized-screenshot.png" ]]; then
+              NOEXTENSION="${NAME%%.png}"
+              # remove alt extension if present
+              ALT="${NOEXTENSION##*.}"
+              ALTTESTS="|dark|"
+              if [[ "$ALTTESTS" == *"|$ALT|"* ]]; then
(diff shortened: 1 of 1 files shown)


vdiff/action.yml
```diff
@@ -300,7 +300,14 @@ runs:
             mkdir -p "./${TEST_PATH}/golden/${TEST}/${BROWSER}"
             for PNG in `ls "${FAIL_DIR}/${BROWSER}"`; do
               NAME=`basename "${PNG}"`
-              if  [[ ${NAME} != *"-diff.png" ]] && [[ ${NAME} != *"-resized-golden.png" ]] && [[ ${NAME} != *"-resized-screenshot.png" ]]; then
+              NOEXTENSION="${NAME%%.png}"
+              # remove alt extension if present
+              ALT="${NOEXTENSION##*.}"
+              ALTTESTS="|dark|"
```
GZolla: I think this is more scalable than adding extra clauses to the if below

svanherk: I'd also be fine with being a bit more lenient, and assuming _anything_ like this is not a golden and shouldn't be moved:
`*-diff.*.png`, `*-resized-screenshot.*.png` and `*-resized-golden.*.png`

Then we could simply change the `if` to this and not need to adjust it for new modes:
```
if [[ ! "${NAME}" =~ -(diff|resized-golden|resized-screenshot)(\.[^.]+)?\.png$ ]]; then
```

GZolla: Yeah that's much cleaner actually, and it seems to work on the test branch 👍

svanherk (approved):

PR #215, opened 2026-06-10
GAUD-10137: pin semantic release

dlockhart:
Semantic release appears to have accidentally published their "next" version as a patch (oh, the irony). Example of a failing release run because of it:
https://github.com/BrightspaceHypermediaComponents/users/actions/runs/27296558905/job/80630676894

I think [they released this feature branch](https://github.com/semantic-release/semantic-release/compare/master...feat/core-integration).


grant-cleary (approved):
I love it. 😆

PR #224, opened 2026-07-09
GAUD-10297 - Fix vdiff for Dependabot PRs with non-read-only tokens

svanherk:
So this is a fun one - discovered this talking to @GZolla, who said Mango mentioned their Dependabot PR was merged because CI was all green, but a vdiff PR was put up. Turns out, if the actor is Dependabot we were exiting early, before marking the build as failed. Looks like this bug has been here for [three years](https://github.com/BrightspaceUI/actions/pull/132), but I don't think this has come up because:
- If Dependabot has read-only permissions, it fails sooner (this is how most of our repos are set up)
- Most of the package updates that cause vdiff failures aren't handled by Dependabot

But still 😬

### DIFF
--- a/vdiff/action.yml
+++ b/vdiff/action.yml
@@ -517,10 +517,6 @@ runs:
         github-token: ${{ inputs.github-token }}
         script: |
           console.log('\x1b[34mUpdating Commit Status');
-          if (context.actor === 'dependabot[bot]') {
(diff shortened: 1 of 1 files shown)


GZolla (approved):

dlockhart (approved):
Fun indeed!

PR #225, opened 2026-07-21
BUC-7532: Setup configuration to allow prereleases

EvanLemieux:
## Description
We are wanting to get rid of Brightspace/github-prerelease, however a number of our backend services rely on it for our deployment strategy. Allowing prereleases via the semantic-release action would enable us to use it without the extra configuration.

This PR is AI-assisted, so please let me know if anything doesn't make sense.

### DIFF
--- a/semantic-release/create-config.sh
+++ b/semantic-release/create-config.sh
@@ -5,9 +5,9 @@ else
   ASSETS="[\"package.json\"]"
 fi
 if [ "$GITHUB_PRERELEASE" = "true" ]; then
-  GITHUB_PRERELEASE_PLUGIN="\"$GITHUB_PRERELEASE_PLUGIN_PATH\""
+  GITHUB_RELEASE_PLUGIN="\"$GITHUB_PRERELEASE_PLUGIN_PATH\""
 else
-  GITHUB_PRERELEASE_PLUGIN="\"@semantic-release/github\""
+  GITHUB_RELEASE_PLUGIN="\"@semantic-release/github\""
 fi
 cat >$FILE_PATH <<EOL
 {
@@ -21,7 +21,6 @@ cat >$FILE_PATH <<EOL
   ],
   "plugins": [
     "@semantic-release/commit-analyzer",
-    "@semantic-release/github",
     $GITHUB_RELEASE_PLUGIN,
     [
       "@semantic-release/npm",
--- a/semantic-release/create-config.sh
+++ b/semantic-release/create-config.sh
@@ -4,6 +4,11 @@ if [ "$IS_TRACKED" != "" ]; then
 else
   ASSETS="[\"package.json\"]"
 fi
+if [ "$GITHUB_PRERELEASE" = "true" ]; then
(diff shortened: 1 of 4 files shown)


semantic-release/create-config.sh
```diff
@@ -4,6 +4,11 @@ if [ "$IS_TRACKED" != "" ]; then
 else
   ASSETS="[\"package.json\"]"
 fi
+if [ "$GITHUB_PRERELEASE" = "true" ]; then
+  GITHUB_PRERELEASE_PLUGIN="\"$GITHUB_PRERELEASE_PLUGIN_PATH\""
+else
+  GITHUB_PRERELEASE_PLUGIN="\"@semantic-release/github\""
```
dlockhart: Isn't this already a plugin that's included? i.e. won't it now be listed twice?

semantic-release/create-config.sh
```diff
@@ -17,6 +22,7 @@ cat >$FILE_PATH <<EOL
   "plugins": [
     "@semantic-release/commit-analyzer",
     "@semantic-release/github",
```
dlockhart: What specifically does `@semantic-release/github` not support about pre-releases that's preventing you from just using it directly? Like I see it has an `addChannel` that claims to set a pre-release.

semantic-release/create-config.sh
EvanLemieux: I'll take a look and remove it if that's the case.

semantic-release/create-config.sh
EvanLemieux: From what I understand, and I may be incorrect. With that we would need a pre-release channel that is separate from the existing semantic-release versioning scheme we have. So `@semantic-release/github` would give us a release like `5.4.53-master.1` and not `5.4.53`.

dlockhart: Are you able to try that out to confirm? I'd just like to confirm that in the however many years since our original version was created that new options haven't been added to the built-in stuff to allow you to do what you need.

EvanLemieux: Sure, I'll see what I can put together to try out.

dlockhart: Thanks.

If that doesn't end up working, then I think since this kind of pre-release workflow seems to [deviate from what semantic-release themselves deem to be "true semantic-release"](https://github.com/semantic-release/semantic-release/issues/3303#issuecomment-4144860661), what might be best would be create your own separate action in your existing [space-github-actions](https://github.com/Brightspace/space-github-actions/tree/main) repo. That way it can be super simple and only worry about the features from here that you want.

svanherk: Yeah, your action likely wouldn't need [this entire block](https://github.com/BrightspaceUI/actions/blob/main/semantic-release/action.yml#L39-L67), which results in a fairly simple semantic-release wrapper with your plugin always added in.

EvanLemieux: My goal has been to move away from space-github-actions where possible, but yeah I'll concede that this might not align with the ideas of semantic-release as a whole.

